PR #2, opened 2023-07-11
Initial go-tuf-metadata client and CI

rdimitrov:
The following PR adds an initial go-tuf-metadata client CLI. 

**Note:**
I've noticed that some of the flag names do not match what we had in the initial document but decided to stick to what we have now, and later we can update the tool along with the client CLIs.

**Disclaimer:** 
There's also a simple CI config file plus a few make targets added with this PR. It's not a huge change so they should be easily reviewed altogether, but I completely agree if we decide to split it to separate PRs.


jku:
On the changes compared to the initial doc: Sorry, I did not modify the doc -- all of the changes have some kind of reason behind them, but it's been a couple of weeks so I'd have to look again to remind myself.


Now that we're adding some infrastructure I want to clearly document that the clients and the infra is here for two reasons:
* testing the conformance test suite itself
* quick-and-dirty test rig while we change things

and that the real setup for running the suite against an implementation should be that the implementation (like python-tuf) 
* maintains a client wrapper in their code base
* installs the suite in CI and 
* runs the suite with their client wrapper in their CI

basically I want to avoid maintaining a list of go dependencies (and rust dependencies and npm dependencies) for just "test data" ... but I don't think this documentation  needs to be in this same PR.

I think I'll just merge this so we can enable the CI and see how it goes

PR #13, opened 2024-03-20
Changes required to pass on macOS without dirtying the repo

segiddins:


jku (approved):
Looks good to me, thanks. 

Virtualenv is a good idea (and running the python client and test suite in same env should be fine for now: no need to get too fancy)

jku:
We could add a github actions matrix to test on both macos and linux as well... but I'll merge this one now

Issue #10, opened 2023-07-17
Figure out an ergonomic way to verify clients current metadata state


jku:
We want to ensure that what the client considers the current valid metadata matches our opinion of the same.
It should be easy for each unitttest to verify this even multiple times during the test run.

* is it enough to verify that the version of each metadata is what we expect? or should we actually compare the file to the file on the repository? I think maybe the latter...
* there's likely several designs that would work: one idea is a method in our UnitTest based class (that does not exist yet. Assume the class knows about the RepositorySimulator and ClientRunner). That method would get a list of roles that we expect to be in client cache and would then compare for each role the metadata in local cache to the metadata in the RepositorySimulator. Then the check in the test would be e.g. `self.assert_client_cache(["root", "timestamp"])` for a test where client refresh ended after timestamp

jku:
I think maybe we're close to something reasonable:
* assert requests to repository:
   repo.metadata_statistics == [("root", 1), ("timestamp", None)]
   repo.artifact_statistics == [("artifact", ("aa61e...acd81")]
* assert client trusted roles and downloaded artifacts:
    `client.trusted_roles() == [("root", 1)]` or `client.version("root") == 1`
   `client.get_downloaded_target_bytes() == [b"file 1 content", b"file 2 content "]`

not the prettiest API but seems to do the job

PR #16, opened 2024-04-26
Use pytest to run tests

segiddins:


tuf_conformance/conftest.py
```diff
@@ -0,0 +1,35 @@
+import os
+import pytest
+
+from tuf_conformance.client_runner import ClientRunner
+from tuf_conformance.simulator_server import SimulatorServer
+
+def pytest_addoption(parser) -> None:
+    """Add `--entrypoint` and `--skip-signing` flags to CLI."""
```
jku, line 8: --skip-signing seems like a copy-paste mistake

tuf_conformance/conftest.py
```diff
@@ -0,0 +1,35 @@
...
+        required=True,
+        type=str,
+    )
+
+@pytest.fixture
+def server():
+    """
+    Parametrize each test with the server under test.
+    """
+    return SimulatorServer(9001)
```
jku, line 22: the pytest fixture idea sounds good but I think we want a single SimulatorServer: otherwise once we have more than one test we'll need dynamic ports and there will be linear startup costs... So server() should return a singleton.

tuf_conformance/test_basic.py
```diff
@@ -4,7 +4,10 @@
 # system to build this eventually 
 
 import argparse
-from typing import Dict, List
+import json
+import os
+
+import pytest  # type: ignore
```
jku, line 10: what is this ignore for?

jku (changes requested):
Nice, looks good.
* `server()` should return a singleton (or we need to redesign the test server)
* I would maybe suggest removing the `test_basic_init_and_refresh` changes from this PR: I agree with the premise but it looks complicated and I think we can do that better in a way that's reusable for future tests in #10 -- maybe the easy first version is just a new method`ClientRunner.verify_metadata([("root", 1), ("timestamp", 1), ("snapshot", 1), ("targets", 1)])` (to just verify the client cache has the correct metadata with correct version numbers)

tuf_conformance/test_basic.py
segiddins: unsure, copy-pasted from the sigstore conformance test setup

tuf_conformance/conftest.py
segiddins: I changed it to use a `yield` which allows us to clean up the server -- this way there's nothing stopping us from e.g. running tests in parallel, or dealing with clearing the state on the server

jku: yield is nice but the issue I was pointing to was multiple servers: This now starts N http servers for N tests I believe. I don't object to that but I'll mention that it was conscious decision to not do that originally:
* I've seen plenty of issues with that in other test suites that start a server per test, especially when parallelized (think of e.g. http server not being available right away because of a hundred http servers starting at the same time)
* the startup and shutdown costs are very real once there are more tests

I'm happy to merge this though, we can always come back to this later

jku (approved):
I'll merge this: I'm not sure about the server design but if someone is willing to do patches, I'm willing to see where it goes :+1:

Issue #43, opened 2024-07-10
Add configuration for "expected failures"


jku:
A recent PR was merged with failing tests (#42).
* it's understandable that a conformance test run may have failures
* it's not useful if CI just fails by default

Let's add a way for clients to set expected failures (sigstore-conformance has a way to do that although it is a little clumsy)

If this is going to take a while, lets disable the failing tests for now.

jku:
two tests are now disabled waiting for this: search for "Ttest"

AdamKorcz:
If the problem is breaking the CI, does it make sense to mock the assertions when we run the conformance suite in the CI? The conformance tests can fail for two reasons: 1) If the tests are failing or 2) if a client is failing. When we run the conformance tests in the CI, we could return `True` in all conformance tests, so that they run to the end and we can test for 1). If we disable tests entirely, we break the UX slightly for users that want to run the conformance suite for a client at this moment at least. 

I suggest two `make` rules:
1. Run all conformance tests with mock assertions which runs in the CI.
2. Run all conformance tests with correct assertions for those that want to use the conformance suite.

jku:
The important case is  where a separate client project is running the suite, the conformance suite CI is secondary:
* client project wants to be informed of any unexpected failures: these should lead to CI failure
* but known, expected failures should not make the whole test suite fail: fixing conformance issues can take a long time (sometimes there's even disagreement about the validity of the issue) so they will want to mark tests as "expected to fail"


So I think we want the test suite to have an input for expected failures that the client projects can use
* in sigstore-conformance there is a `--xfails` option that takes a list of test names and then the workflows use it like this: https://github.com/sigstore/sigstore-python/blob/main/.github/workflows/conformance.yml
* I'd be fine with that but it is a little clunky. I suppose --xfails could take a file name as well, and then the file contains all expected to fail test names

PR #45, opened 2024-07-10
disable two failing conformance tests

AdamKorcz:
This will prevent two failing conformance tests from running in the CI.


jku (approved):
thanks, let's go with this until #43 exists

Issue #35, opened 2024-07-04
Consider github action


jku:
sigstore-conformance provides a GitHub action that client projects can use. A similar setup might work for us: this would also mean we don't necessarily need to setup pypi releases etc...

This is not needed right now but at some point.

jku:
> This is not needed right now but at some point

After discussing with Lukas: maybe this is actually a priority...
* if we add an action now, we pretty much have a MVP that we can show to client projects
* we can even add a workflow to python-tuf as an example

As long as we make it clear that the "client wrapper definition" is not stable yet this should be fine

jku:
I've implemented an initial version. Further plans:
* Move a client wrapper and workflow to an external repo to test end-to-end and to have demonstration -- I think I'll add this to python-tuf (but I'll leave the test trigger to manual dispatch only so we can keep breaking things in tuf-conformance if needed)
* Add support for things like expected failures: #43
* Polish the output and debugging information #32
  * maybe upload an artifact for each failing test with the repository metadata and client output?

AdamKorcz:
I suggest we add some basic documentation on adopting the GH action.

jku:
I think everything here has been acted on or individual tickets have been filed

PR #46, opened 2024-07-10
Add a simple action to run the tests suite

jku:
* Modify the CI workflow so it tests the action
* The action does not use the Makefile since that felt quite awkward

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -1,0 +1,27 @@
+name: "tuf-conformance"
+description: "TUF client conformance test suite"
+
+inputs:
+  entrypoint:
+    description: "client wrapper to invoke"
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - name: Set up Python
+      uses: actions/setup-python@82c7e631bb3cdc910f68e0081d67478d79c6982d # v5.1.0
+      with:
+        python-version: "3.11"
+        cache: "pip"
+
+    - name: Install test
+      run: |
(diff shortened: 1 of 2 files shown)


jku:
Bit of self-review:
* The action does not use Makefile: this is annoying in the sense that we now have two separate ways of installing and executing the suite
* I did this because input arguments and Makefiles don't feel like a great combo. Currently we only have  --entrypoint but there will be more
* that said, the action still looks clean and I suppose the Makefile can remain a development aid. I think this is acceptable

This was quite quick to whip up: I'm happy to try different ideas if you have any @AdamKorcz, @DavidKorczynski  but I think this should work. I'll document some further plans in #35

action.yml
```diff
@@ -0,0 +1,27 @@
...
+
+    - name: Install test
+      run: |
+        pip install -e "${{ github.action_path }}"
+      shell: bash
+
+    - name: Run test suite
+      id: sigstore-conformance
+      run: |
+        pytest tuf_conformance --entrypoint "${{ inputs.entrypoint }}"
```
jku, line 26: this input really should get handled via an environment variable to prevent injection

AdamKorcz:
I think not invoking the Makefile here is fine. I don't expect the way we run the test suite to change often, so it shouldn't be a lot of maintenance.

AdamKorcz (approved):

Issue #39, opened 2024-07-08
separate self tests from actual test suite?


jku:
Currently the "self tests" (tests that verify the test suite components work correctly)  and actual conformance tests are in the same directory and they all get executed at once
* this is probably incorrect: we should have separate runs for test suite and and "self tests" (currently test_repository_simulator.py) 
* it's also confusing for someone trying to find the actual conformance tests

opinions @AdamKorcz ?

AdamKorcz:
I think we should have

1. a `make` rule that runs all test
2. a `make` rule that runs only conformance tests
3. a `make` rule that runs only "self tests"

IMO #3 should run in the CI of the conformance test suite.

PR #47, opened 2024-07-10
Move self-tests to their own directory

AdamKorcz:
This will prevent the self-tests when running the conformance tests.

Closes https://github.com/theupdateframework/tuf-conformance/issues/39


jku (approved):
LGTM: makes sense to not run these as part of conformance tests. 

Do you want to add a self-test job to our CI workflow so we'll still have these tests running in CI?

AdamKorcz:
> LGTM: makes sense to not run these as part of conformance tests.
> 
> Do you want to add a self-test job to our CI workflow so we'll still have these tests running in CI?

Yep, that would be great. Let's merge this in for now and I will follow up with that.

PR #48, opened 2024-07-10
clients: Tweak python client

jku:
This is a result of linting the code when adding a copy to python-tuf. The changes look good to me and should not include runtime changes.


jku:
Making this a draft: I might need to add another tweak

AdamKorcz (approved):

PR #50, opened 2024-07-10
action: Fix the test path

### LINKED ISSUE #49: action expects to be running in the tuf-conformance source dir
The action I added works when tested as a local action (`uses: ./`) but fails when it's used externally:

This is likely because in the command `pytest tuf_conformance --entrypoint "$ENTRYPOINT"` _tuf_conformance_ is a directory that is part of the tuf-conformance suite -- possibly it just needs to be prefixed with `${{ github.action_path }}`

jku:
When the action is used from an external project, the test path (_tuf_conformance_) is not in working dir: prefix the the test path with action location

Fixes #49

### DIFF
--- a/action.yml
+++ b/action.yml
@@ -24,6 +24,7 @@ runs:
       id: sigstore-conformance
       env:
         ENTRYPOINT: ${{ inputs.entrypoint }}
+        TEST_LOCATION: ${{ github.action_path }}/tuf_conformance
       run: |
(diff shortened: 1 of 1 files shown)


jku:
I tested this in https://github.com/jku/python-tuf/actions/runs/9874869528/job/27270216176 (unsure what the visibility of that is but it's a successful conformance run in  a python-tuf repo)

AdamKorcz (approved):

PR #67, opened 2024-07-15
Add two tests for custom fields

AdamKorcz:
@jku Could you check `test_custom_field_in_meta` for correctness? Essentially, this test bumps the repo metadata so the client should update. It then adds a custom field to `TimestampMetadata["signed"]["meta"]["snapshot.json"]`: `["hash"] = "custom_field_value"`. Then the client updates and the test expects that the update should fail because there are no signatures that will verify `TimestampMetadata["signed"]`. 

My main concerns are:
1. Is `TimestampMetadata["signed"]["meta"]["snapshot.json"]["hash"]` a valid custom field according to the TUF specification? 
2. Should the client include `TimestampMetadata["signed"]["meta"]["snapshot.json"]["hash"]` when verifying keys?
3. Should the client store `TimestampMetadata["signed"]["meta"]["snapshot.json"]["hash"]` in its local metadata?

Right now the test fails at this assertion:

```python
    with open(os.path.join(client.metadata_dir, "timestamp.json"), "r") as client_timestamp_json:
        print("client timestamp: ", client_timestamp_json)
        assert json.load(client_timestamp_json)[0]["signed"]["meta"]["snapshot.json"]["hash"] == "custom_field_value"
```

... and when checking the local snapshot and timestamp version. The client bumps to version 2 for both and the test expects that it doesn't.

```python
    assert client._version_equals(Root.type, 2)
    assert client._version_equals(Snapshot.type, 1)
    assert client._version_equals(Timestamp.type, 1)
```

### DIFF
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -259,6 +259,8 @@ class RepositorySimulator():
         # update repo metadata
         self.save_metadata_bytes(role, new_ss_bytes)
         md_bytes = self.load_metadata_bytes(role)
+
+        # Verify the json is valid
         ss_obj = json.loads(md_bytes)
 
     def add_key(self, delegator: str, role: str, signer: Signer) -> None:
--- a/tuf_conformance/test_basic.py
+++ b/tuf_conformance/test_basic.py
@@ -8,11 +8,53 @@ import copy
 from tuf_conformance.repository_simulator import RepositorySimulator
 from tuf_conformance.simulator_server import SimulatorServer
 from tuf_conformance.client_runner import ClientRunner
+from tuf_conformance.utils import meta_dict_to_bytes
 
 from tuf.api.metadata import (
     Timestamp, Snapshot, Root, Targets, Metadata
 )
 
+def test_custom_field_hash_verificatio(client: ClientRunner,
+                            server: SimulatorServer) -> None:
+    """Tests that custom fields are included in hash verification"""
+
+    name = "test_custom_field_hash_verification"
+
+    # initialize a simulator with repository content we need
+    repo = RepositorySimulator()
+    server.repos[name] = repo
+    init_data = server.get_client_init_data(name)
+    assert client.init_client(init_data) == 0
+    client.refresh(init_data)
+    # Sanity check
+    assert client._files_exist([Root.type,
+                               Timestamp.type,
+                               Snapshot.type,
+                               Targets.type])
+    # Bump so that client should update
+    repo.bump_version_by_one(Snapshot.type) # v2
+
+    # Add a custom field without re-calculating hashes
+    new_snapshot_md = json.loads(repo.md_snapshot_json)
+    new_snapshot_md["signed"]["custom_field"] = "custom_field_value"
+    repo.save_metadata_bytes(Snapshot.type, meta_dict_to_bytes(new_snapshot_md))
+
+    # Client updates
+    client.refresh(init_data)
+
+    # Client should not bump to v2 because the repo metadata
+    # has not recalculated the hashes
+    assert client._version_equals(Snapshot.type, 1)
+
+    # Update and also recalculate hashes (update_snapshot
+    # recalculates lengths and hashes)
+    repo.update_snapshot()
+
+    # Client updates
+    client.refresh(init_data)
+
+    # Now the client should update properly
+    assert client._version_equals(Snapshot.type, 3)
 
 
 def test_TestTimestampEqVersionsCheck(client: ClientRunner, server: SimulatorServer) -> None:
--- a/tuf_conformance/test_keys.py
+++ b/tuf_conformance/test_keys.py
@@ -323,3 +323,66 @@ def test_duplicate_keys_root(client: ClientRunner,
     assert len(md_root.roles[Snapshot.type].keyids) == 1
     # TODO: Double check that "1" is correct here:
     assert len(md_snapshot.signatures) == 1
+
+
+def test_custom_field_in_meta(client: ClientRunner,
+                            server: SimulatorServer) -> None:
+    """Bumps up the version at adds a custom field in the timestamp
+    metadata at MD["signed"]["meta"]["snapshot.json"]["hash"].
+    Then the client bumps and the test checks for the expected
+    metadata version."""
+
+    name = "test_custom_field_in_meta"
+
+    # initialize a simulator with repository content we need
+    repo = RepositorySimulator()
+    server.repos[name] = repo
+    init_data = server.get_client_init_data(name)
+    assert client.init_client(init_data) == 0
+    client.refresh(init_data)
+    # Sanity check
+    assert client._files_exist([Root.type,
+                               Timestamp.type,
+                               Snapshot.type,
+                               Targets.type])
+
+    new_root_md = json.loads(repo.md_root_json)
+    new_root_md["signed"]["roles"]["timestamp"]["threshold"] = 2
+    repo.save_metadata_bytes(Root.type, meta_dict_to_bytes(new_root_md))
+
+    # Add a valid key and bump
+    repo.add_key_to_role(Timestamp.type)
+    # Bump so that client should update
+    repo.update_snapshot() # snapshot: v2, timestamp: v2
+    repo.bump_root_by_one() # v2
+    assert len(json.loads(repo.md_root_json)["signed"]["roles"]["timestamp"]["keyids"]) == 2
+
+    # Add a custom field without re-calculating hashes
+    new_timestamp_md = json.loads(repo.md_timestamp_json)
+    # "hash" is a custom field
+    new_timestamp_md["signed"]["meta"]["snapshot.json"]["hash"] = "custom_field_value"
+    repo.save_metadata_bytes(Timestamp.type, meta_dict_to_bytes(new_timestamp_md))
+    assert client._version_equals(Timestamp.type, 1)
+
+    # Client updates
+    client.refresh(init_data)
+    with open(os.path.join(client.metadata_dir, "timestamp.json"), "r") as client_timestamp_json:
+        print("client timestamp: ", client_timestamp_json)
+        assert json.load(client_timestamp_json)[0]["signed"]["meta"]["snapshot.json"]["hash"] == "custom_field_value"
+
+    # Client should not bump snapshot to v2 because the repo metadata
+    # has not recalculated the hashes
+    assert client._version_equals(Root.type, 2)
+    assert client._version_equals(Snapshot.type, 1)
+    assert client._version_equals(Timestamp.type, 1)
+
+    # Update and also recalculate hashes (update_snapshot
+    # recalculates lengths and hashes)
+    repo.update_snapshot() # timestamp: v3, snapshot: v3
+
+    # Client updates
+    client.refresh(init_data)
+
+    # Now the client should update properly
+    assert client._version_equals(Snapshot.type, 3)
+    assert client._version_equals(Timestamp.type, 3)


jku:
The field in your example seems legal.

The tests feels very complicated for what I think it's doing and I can't quite follow it (and have trouble using the RepositorySimulator so can't give you a better example). With python-tufs repositorySimulator I created this test for the same purpose:

```
    def test_unrecognised_fields(self) -> None:
        self.sim.timestamp.snapshot_meta.unrecognized_fields["hash"] = "value"
        self._run_refresh()

        timestamp_path = os.path.join(self.metadata_dir, "timestamp.json")
        timestamp: Metadata[Timestamp] = Metadata.from_file(timestamp_path)
        self.assertEqual(
            "value", timestamp.signed.snapshot_meta.unrecognized_fields["new_field"]
        )
```

In other words we just want to verify that the client includes the new field when it calculates the signatures over timestamp: in this case it does since the update succeeded.

AdamKorcz:
Thanks, @jku `test_custom_field_in_meta` is not exactly the same as `test_unrecognised_fields` from python-tuf. `test_unrecognised_fields` does the following:

1. The repo set a threshold of 2 keys.
2. The repo adds one more key to a total of 2.
3. The repo adds the custom field in `snapshot_meta`. The repository does not add more keys, so I expect that the two keys that were added prior should not be valid now. 
4. The client updates and I expect it to fail because there are no keys that have signed the current timestamp metadata in the repo. Currently the client succeeds, but it does not store the custom field added in #3 above; it fails on line 371.

jku:
> 2. The repo adds one more key to a total of 2.
> 3. The repo adds the custom field in the timestamps snapshot_meta. The repository does not add more keys, so I expect that the two keys that were added prior should not be valid now.

I think there is some misunderstanding here. I don't understand why a key change is needed here, or why the timestamp metadata  would be invalid if the timestamp role has two keys

AdamKorcz:
> I think there is some misunderstanding here. I don't understand why a key change is needed here, or why the timestamp metadata would be invalid if the timestamp role has two keys

The reason would be that:
1. we modify `Timestamp.signed` _after_ we have added the keys and after we have signed `Timestamp.signed`; and
2. we don't add new keys or sign again

### State after step 2

After step 2 above, the root metadata in the repository looks like this:

```json
{
  "signatures": [
    {
      "keyid": "f28172840549b4dd92af1b79f25ac6f44d9258cea735d0cd4f9cc60b6a14ab20",
      "sig": "30450220453a843234119e51535c93d7fd96d71567517e76ff9402583494b8f5299f1a1a022100f48cd5f9ebadd531741498f071fb753c2e5802fc5c4b8da25e3a5c4eb66acb56"
    }
  ],
  "signed": {
    "_type": "root",
    "consistent_snapshot": true,
    "expires": "2024-08-15T15:58:12Z",
    "keys": {
      "1080d3f217889cf21fe393f1f2ba8e93a18f7d2c8f58c316ef715e4025a3ca22": {
        "keytype": "ecdsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEl7BmVfytSZqf+daGwQYUyGrenUIj\\niMogTvCCI82qh3lySEbNGuZ/YVt2Z4BAubWTcFlm5w/iX9ivYmBHWa2Utg==\\n-----END PUBLIC KEY-----\\n"
        },
        "scheme": "ecdsa-sha2-nistp256"
      },
      "53ae57d05515d78a89dc18c249d00153f436a1ce94c4da52e829e0c61f5b398b": {
        "keytype": "ecdsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEETonewvdNKkH21CIXQKKenDD07Zf\\nGTbxOfSDEwtqzhIWLRBw7GTXEy44Wksh8cda08nooqDo50OPqYwUoETPYw==\\n-----END PUBLIC KEY-----\\n"
        },
        "scheme": "ecdsa-sha2-nistp256"
      },
      "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d": {
        "keytype": "ecdsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEfx0l4o9d6o7z1daoU0zQJN/l10SM\\ntvYNGr/AlQIESQ4bDOOs02f4RrMmIPvwp1x9NcvFvJ/r0AGOutFeHJVqyA==\\n-----END PUBLIC KEY-----\\n"
        },
        "scheme": "ecdsa-sha2-nistp256"
      },
      "dee9b348f286cb1c3934162450e6f1e33c691be6d6cb508af3fb099451b00f6d": {
        "keytype": "ecdsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAE5yxYsEKJ84xYc1p6AOf5x32kUOKY\\nHLAncIupxBDT6H0WlMFBt1JAW6vcM2XO2HKfpp7lcxJPvEhe9jLL5j7Ddw==\\n-----END PUBLIC KEY-----\\n"
        },
        "scheme": "ecdsa-sha2-nistp256"
      },
      "f28172840549b4dd92af1b79f25ac6f44d9258cea735d0cd4f9cc60b6a14ab20": {
        "keytype": "ecdsa",
        "keyval": {
          "public": "-----BEGIN PUBLIC KEY-----\\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEhmkARbOKij3n/D4VDxbx2SnjkjDe\\n59GFlUGDb4fzCgI9PRWLhrrHcMjHQr5CQO+I5aKGM8UjmF8RWd+6B+2ktA==\\n-----END PUBLIC KEY-----\\n"
        },
        "scheme": "ecdsa-sha2-nistp256"
      }
    },
    "roles": {
      "root": {
        "keyids": [
          "f28172840549b4dd92af1b79f25ac6f44d9258cea735d0cd4f9cc60b6a14ab20"
        ],
        "threshold": 1
      },
      "snapshot": {
        "keyids": [
          "53ae57d05515d78a89dc18c249d00153f436a1ce94c4da52e829e0c61f5b398b"
        ],
        "threshold": 1
      },
      "targets": {
        "keyids": [
          "dee9b348f286cb1c3934162450e6f1e33c691be6d6cb508af3fb099451b00f6d"
        ],
        "threshold": 1
      },
      "timestamp": {
        "keyids": [
          "1080d3f217889cf21fe393f1f2ba8e93a18f7d2c8f58c316ef715e4025a3ca22",
          "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d"
        ],
        "threshold": 2
      }
    },
    "spec_version": "1.0.31",
    "version": 2
  }
}
```

The important part is this:

```json
      "timestamp": {
        "keyids": [
          "1080d3f217889cf21fe393f1f2ba8e93a18f7d2c8f58c316ef715e4025a3ca22",
          "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d"
        ],
        "threshold": 2
      }
```

And the timestamp metadata looks like this:

```json

{
  "signatures": [
    {
      "keyid": "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d",
      "sig": "30460221009e9f404882439c5e512815612defd8a839548127f9a13aca2832f948b2432612022100b2e4c035705ce329016a0693537ae3fa787ae0eab6a8ad3d4ec0b5f750cc9dfb"
    }
  ],
  "signed": {
    "_type": "timestamp",
    "expires": "2024-08-15T15:58:12Z",
    "meta": {
      "snapshot.json": {
        "version": 2
      }
    },
    "spec_version": "1.0.31",
    "version": 2
  }
}
```

### State after step 3

After step 3, the root metadata in the repo remains the same, and the timestamp metadata looks like this:

```json
{
  "signatures": [
    {
      "keyid": "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d",
      "sig": "30460221009e9f404882439c5e512815612defd8a839548127f9a13aca2832f948b2432612022100b2e4c035705ce329016a0693537ae3fa787ae0eab6a8ad3d4ec0b5f750cc9dfb"
    }
  ],
  "signed": {
    "_type": "timestamp",
    "expires": "2024-08-15T15:58:12Z",
    "meta": {
      "snapshot.json": {
        "hash": "custom_field_value",
        "version": 2
      }
    },
    "spec_version": "1.0.31",
    "version": 2
  }
}
```

As such, the keys for the timestamp in the root metadata remain the same:

```json
      "timestamp": {
        "keyids": [
          "1080d3f217889cf21fe393f1f2ba8e93a18f7d2c8f58c316ef715e4025a3ca22",
          "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d"
        ],
        "threshold": 2
      }
```

... but the signed part in the timestamp metadata is different:

```json
{
  "signatures": [
    {
      "keyid": "84e670f2f337398b12057e79223bc7ebd8a0a79c57deeb760b878b8dfe64877d",
      "sig": "30460221009e9f404882439c5e512815612defd8a839548127f9a13aca2832f948b2432612022100b2e4c035705ce329016a0693537ae3fa787ae0eab6a8ad3d4ec0b5f750cc9dfb"
    }
  ],
  "signed": {
    "_type": "timestamp",
    "expires": "2024-08-15T15:58:12Z",
    "meta": {
      "snapshot.json": {
        "hash": "custom_field_value", <------ This part has not been signed by the keys in the root metadata
        "version": 2
      }
    },
    "spec_version": "1.0.31",
    "version": 2
  }
}
```

What is the expected behaviour when the client now updates?

jku:
So:
* it is clear that the timestamp metadata after step 3 is invalid as it has has signature threshold has not been reached
* it is still unclear how this is related to testing if custom fields are being correctly included in the signature verification?

In other words, I still do not see any reason to do key changes in this test.

---

`assert json.load(client_timestamp_json)[0]["signed"]["meta"]["snapshot.json"]["hash"] == "custom_field_value"`

this seems like an incorrect assert: `json.load(client_timestamp_json)[0]` is not valid as far as I can tell. Have I mentioned I think unnecessary json wrangling is not a great pattern :) 

--- 

I will have a look at why the client does not fail the refresh as you expect -- but as said I don't think this is in any way related to custom fields (you can take out the custom field editing and the test still fails)

jku:
I believe the timestamp is signed correctly:
```
 "signatures": [
  {
   "keyid": "e030441141fce170ca5caf5c893c3c4aa8d1931309fce4b2ba4edf3d010cde47",
   "sig": "304502204293f313de714b9236c6c6cd84f53e29d9691fdc729ea508b7b0de0e95771da702210095887222312267c6252cf222ba765c290497a8554190d56630833fc170f1f2d9"
  },
  {
   "keyid": "e4adbf7a5e2fe7b0b4ef998dbf48a3832b8206ac90c9a2674c18eceab8c203b6",
   "sig": "304502200bf3f31253d59620959659d29526ab090400e184a35a5912459c947767223b7f022100ac50f568451f3e82452b84573a03d1af0af486c9c1a69040f0375ea4cd744c24"
  }
 ],
 "signed": {
  "_type": "timestamp",
  "expires": "2024-08-16T07:51:38Z",
  "meta": {
   "snapshot.json": {
    "hash": "custom_field_value",
    "version": 2
   }
  },
  "spec_version": "1.0.31",
  "version": 2
 }
}

```

jku:
I've added a simple custom field test in #92. That one does not do any of the threshold and key changes that are done here as I don't think they're relevant to the custom fields themselves.

I think we can close this -- unless you think there is a test we should add for the keys/threshold change here.

jku:
closing based on comments above and chat yesterday

PR #87, opened 2024-07-27
Add test for threshold

AdamKorcz:
Adds a similar test to https://github.com/theupdateframework/tuf-conformance/pull/86 but with a threshold. Our two integrated clients have an inconsistency here it looks like. The root cause may be identical to https://github.com/theupdateframework/tuf-conformance/pull/86, but go-tuf is throwing a different error, so it might be interesting to consider both.

### DIFF
--- a/tuf_conformance/test_keys.py
+++ b/tuf_conformance/test_keys.py
@@ -37,6 +37,34 @@ def initial_setup_for_key_threshold(client: ClientRunner,
     assert client._version(Root.type) == 4
 
 
+def test_snapshot_threshold_simple(
+    client: ClientRunner,
+    server: SimulatorServer
+) -> None:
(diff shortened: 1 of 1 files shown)


jku:
similar root cause seems likely: the actual public key is not in the metadata. `repo.root.add_key()` should take care of both adding the keyid to `root.roles[role].keyids` and putting the key content into `root.keys`: that should be fine here.

jku:
#144 is for handling the keyid-without-key case: this test should be rewritten to use add_key() instead of modifying keyids directly.

jku (changes requested):
marking "request changes", see previous comment

jku:
I believe the tests in test_updater_key_rotations.py should cover this case. Closing

PR #89, opened 2024-07-30
simplify test setup

jku:
* get rid of string names for tests: use request.node.originalname
* Do the repository and client data setup in one call to remove 3 lines of boiler plate from each test
* standardize whitespace use in function argument list (this should be compatible with code formatters)

Note: I'm pretty sure we could avoid the RequestFixture argument to every test by being clever in the server fixture... but the pytest fixtures are so magical to me that I decided to keep the usage to the simplest possible even if that meant more changed lines in this PR

### DIFF
--- a/tuf_conformance/simulator_server.py
+++ b/tuf_conformance/simulator_server.py
@@ -54,8 +54,17 @@ class SimulatorServer(ThreadingHTTPServer):
         # key is test name, value is the repository sim for that test
         self.repos: Dict[str, RepositorySimulator] = {}
 
-    def get_client_init_data(self, repo: str) -> ClientInitData:
-        return ClientInitData(
-            f"http://{self.server_address[0]}:{self.server_address[1]}/{repo}/metadata/",
-            self.repos[repo].fetch_metadata("root", 1)
+    def new_test(self, name: str) -> tuple[ClientInitData, RepositorySimulator]:
+        """Return a tuple of
+        * A new repository simulator (for test case to control)
+        * client initialization parameters (so client can find the simulated repo)
+        """
+        repo = RepositorySimulator()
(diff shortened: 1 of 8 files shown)


jku:
> Note: I'm pretty sure we could avoid the RequestFixture argument to every test by being clever in the server fixture... but the pytest fixtures are so magical to me that I decided to keep the usage to the simplest possible even if that meant more changed lines in this PR

actually there might be a clean way to do this that would be useful for the repository dump feature as well: I will try

jku:
> * get rid of string names for tests: use request.node.originalname

it's possible that this turns out to be the wrong choice when e.g. parametrized tests are needed: I'm sure there is a useful name for that situation too, we can fix change the choice then

AdamKorcz (approved):

tuf_conformance/simulator_server.py
```diff
@@ -54,8 +54,17 @@ def __init__(self):
...
+    def new_test(self, name: str) -> tuple[ClientInitData, RepositorySimulator]:
+        """Return a tuple of
+        * A new repository simulator (for test case to control)
+        * client initialization parameters (so client can find the simulated repo)
+        """
+        repo = RepositorySimulator()
+        self.repos[name] = repo
+
+        client_data = ClientInitData(
+            f"http://{self.server_address[0]}:{self.server_address[1]}/{name}/metadata/",
```
segiddins, line 66: should this be `parse.quote(name)`? Ruby's `URI` complains about unescaped brackets `[]` in the URI

jku: you are right and then it needs to be unquoted when the request is handled (I'll take care of that in #98)

PR #92, opened 2024-07-31
Custom field tests

### LINKED ISSUE #28: tests: check for support of custom fields
* spec requires support custom fields anywhere in the json :
    > Implementers who encounter undefined attribute-value pairs in the format must include the data when calculating hashes or verifying signatures and must preserve the data when re-serializing
* we should test this. python-tuf has a unrecognized_fields in (all?) it's objects that can be used for this
* Mostly the test is interesting if the custom fields are withing "signed" (to ensure hash calculation contains these fields)

jku:
Add simplified test for custom fields.

 Fixes #28. See also #67


jku:
rebased on main, ready for review

AdamKorcz (approved):

Issue #100, opened 2024-08-02
Fix TODOs in tests or remove incomplete tests


jku:
example: it's not obvious what `test_TestDelegateConsistentSnapshotDisabled()` gives us: it seems to be lacking both a description of what it is testing and the HTTP request assertions that the upstream go-tuf test has.

Let's go through the tests with TODOs and FIXMEs and either complete them or remove them as incomplete.

AdamKorcz:
Removing `test_TestDelegateConsistentSnapshotDisabled` in https://github.com/theupdateframework/tuf-conformance/pull/103

jku:
Filed #107 as a specific sub-issue

AdamKorcz:
I believe all TODOs in tests are resolved.

PR #102, opened 2024-08-02
add comments to file download test

AdamKorcz:
Ongoing work on https://github.com/theupdateframework/tuf-conformance/issues/100

### DIFF
--- a/tuf_conformance/test_file_download.py
+++ b/tuf_conformance/test_file_download.py
@@ -1,7 +1,6 @@
 import os
 from tuf.api.metadata import Snapshot
 
-from tuf_conformance.repository_simulator import RepositorySimulator
 from tuf_conformance.simulator_server import SimulatorServer
 from tuf_conformance.client_runner import ClientRunner
 from tuf_conformance import utils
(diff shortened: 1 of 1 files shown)


tuf_conformance/test_file_download.py
```diff
@@ -268,74 +310,93 @@ def test_multiple_changes_to_target(
...
-        # at the same path as the legitimate one.
+        # The repository now creates a malicious file on with the
+        # same name a the legitimate file and bumps the targets
+        # and root metadata.
         malicious_file_contents_str = f"malicious-file-contents-{i}"
         new_target_file = open(target_file_path, 'w+')
         new_target_file.write(malicious_file_contents_str)
         new_target_file.close()
+        repo.targets.version += 1
+        repo.bump_root_by_one()
```
jku, line 390: these sort of changes seem a little strange:
* the root change affects nothing in the test as far as I can see
* the targets version change is a bit iffy: repository now serves a targets N+1 but it's not actually part of snapshot so likely no-one ever tries to download that -- the targets version is also bumped earlier in the loop (and snapshot is updated)

jku (approved):
This looks fine. 

I filed another issue about reviewing this (#107 ) but I can take a stab at that at some point after this PR: I think these tests are more complicated than they need to be, I'm pretty sure none of this file writing needs to be done by the test itself: `RepositorySimulator.add_target()` should work and if it does not we can fix it

PR #104, opened 2024-08-03
add test for non-root rotations

AdamKorcz:
Migrates the `test_non_root_rotations` test from python-tuf.

### DIFF
--- a/tuf_conformance/test_updater_key_rotations.py
+++ b/tuf_conformance/test_updater_key_rotations.py
@@ -217,11 +217,10 @@ def test_non_root_rotations(client: ClientRunner,
         repo.publish_root()
 
         # run client workflow, assert success/failure
-        expected_error = md_version.res
-        if expected_error:
+        expected_result = md_version.res
+        if expected_result:
             assert client.refresh(init_data) == 0
 
-            # Call fetch_metadata to sign metadata with new keys
             expected_local_md: bytes = repo.fetch_metadata(role)
             # assert local metadata role is on disk as expected
             md_path = os.path.join(client.metadata_dir, f"{role}.json")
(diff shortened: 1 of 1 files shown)


tuf_conformance/test_updater_key_rotations.py
```diff
@@ -145,3 +173,61 @@ def test_root_rotation(client: ClientRunner,
...
+
+        repo.root.version += 1
+        repo.publish_root()
+
+        # run client workflow, assert success/failure
+        expected_error = md_version.res
+        if expected_error:
+            assert client.refresh(init_data) == 0
+
+            # Call fetch_metadata to sign metadata with new keys
```
jku, line 224: nit: this comment is likely outdated, we call fetch_metadata() here to be able to do the comparison below not specifically to sign

tuf_conformance/test_updater_key_rotations.py
```diff
@@ -145,3 +173,61 @@ def test_root_rotation(client: ClientRunner,
...
+
+        for i in md_version.sigs:
+            repo.add_signer(role, signers[i])
+
+        repo.root.version += 1
+        repo.publish_root()
+
+        # run client workflow, assert success/failure
+        expected_error = md_version.res
+        if expected_error:
```
jku, line 221: naming is not great: when the test is expecting success, `expected_error==True`

jku (approved):
looks pretty good, thanks. There is a variable name that could still be fixed but I'm fine with this as well

tuf_conformance/test_updater_key_rotations.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/104/commits/bdc11da0cac16015bcca384c369e94c58ee36ef4

tuf_conformance/test_updater_key_rotations.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/104/commits/bdc11da0cac16015bcca384c369e94c58ee36ef4

PR #108, opened 2024-08-05
Add test for duplicate keyids

AdamKorcz:
Add test that is specified in https://github.com/theupdateframework/tuf-conformance/pull/86#issuecomment-2259995260 without the bonus test.

The two clients seem to disagree on whether root metadata should update if there are duplicate keys. As such, python-tuf does not fail on meeting the threshold, but rather because the root has duplicate keys.

### DIFF
--- a/tuf_conformance/test_keys.py
+++ b/tuf_conformance/test_keys.py
@@ -199,3 +199,30 @@ def test_duplicate_keys_root(
     # The clients snapshot metadata should still
     # be version 1
     assert client.version(Snapshot.type) == 1
+
+
+def test_duplicate_public_key(
+    client: ClientRunner, server: SimulatorServer
+) -> None:
+    """Tests whether clients can consider the same
+    signature twice when there is a threshold"""
+    init_data, repo = server.new_test(client.test_name)
+
+    assert client.init_client(init_data) == 0
+    client.refresh(init_data)
+    repo.bump_root_by_one()  # v2
+
+    assert len(repo.root.roles["snapshot"].keyids) == 1
+    assert len(repo.md_snapshot.signatures) == 1
+    snapshot_keys = repo.md_snapshot.signatures
+    first_snapshot_keyid = next(iter(snapshot_keys))
+    repo.root.roles[Snapshot.type].keyids.append(first_snapshot_keyid)
+    repo.bump_root_by_one()  # v3
+    client.refresh(init_data)
+    repo.root.roles[Snapshot.type].threshold = 2
+    repo.bump_root_by_one()  # v4
+    repo.update_snapshot()  # v2
+    client.refresh(init_data)
+
+    assert client.version(Snapshot.type) == 1
+    assert client.version(Root.type) == 4
--- a/tuf_conformance/client_runner.py
+++ b/tuf_conformance/client_runner.py
@@ -7,7 +7,7 @@ from typing import Iterable, Optional
 from tuf_conformance.simulator_server import ClientInitData, SimulatorServer
 
 from tuf_conformance.metadata import (
-    MetadataTest
+    MetadataTest, JSONDeserializerTest
 )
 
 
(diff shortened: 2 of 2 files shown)


tuf_conformance/test_keys.py
```diff
@@ -199,3 +199,30 @@ def test_duplicate_keys_root(
...
+    assert len(repo.md_snapshot.signatures) == 1
+    snapshot_keys = repo.md_snapshot.signatures
+    first_snapshot_keyid = next(iter(snapshot_keys))
+    repo.root.roles[Snapshot.type].keyids.append(first_snapshot_keyid)
+    repo.bump_root_by_one()  # v3
+    client.refresh(init_data)
+    repo.root.roles[Snapshot.type].threshold = 2
+    repo.bump_root_by_one()  # v4
+    repo.update_snapshot()  # v2
+    client.refresh(init_data)
```
jku, line 225: asserts for success/failure on the refresh calls would make it more readable

tuf_conformance/test_keys.py
```diff
@@ -199,3 +199,30 @@ def test_duplicate_keys_root(
...
+def test_duplicate_public_key(
+    client: ClientRunner, server: SimulatorServer
+) -> None:
+    """Tests whether clients can consider the same
+    signature twice when there is a threshold"""
+    init_data, repo = server.new_test(client.test_name)
+
+    assert client.init_client(init_data) == 0
+    client.refresh(init_data)
+    repo.bump_root_by_one()  # v2
```
jku, line 213: nit: this call seems irrelevant to the test, or am I missing something?

jku (commented):
The code comments are nits, the return value might be useful to check mostly as documentation to test reader: otherwise it's hard to figure out what is actually expected

> The two clients seem to disagree on whether root metadata should update if there are duplicate keys. As such, python-tuf does not fail on meeting the threshold, but rather because the root has duplicate keys. 

Can you be more specific, I can't find code that would fail on duplicates in  `root.roles["role"].keyids`? Do you mean signatures?

jku (approved):

jku:
> Can you be more specific, I can't find code that would fail on duplicates in `root.roles["role"].keyids`? Do you mean signatures?

oh I see now, you mean the test failing in CI. I'll have a look

AdamKorcz:
> > Can you be more specific, I can't find code that would fail on duplicates in `root.roles["role"].keyids`? Do you mean signatures?
> 
> oh I see now, you mean the test failing in CI. I'll have a look
> 
> EDIT:
> 
> ```
>         if len(set(keyids)) != len(keyids):
>             raise ValueError(f"Nonunique keyids: {keyids}")
> ```
> 
> python tuf indeed will not deserialize metadata with duplicate keyids... we'll have to decide how we continue. This still seems like an acceptable choice by python-tuf: I wouldn't want to force them to change for the test...

Yes, for the record, the test is currently set up so that python-tuf fails, but that does not mean that it - or go-tuf for that matter - is wrong. The interesting part here is that the two clients behave differently.

jku:
a way to move forward here would be to accept either case for now:
* failure to refresh metadata with duplicate keyids -> **Test succeeds**
* metadata with duplicate keyids works but keyid is not double counted for threshold -> **Test succeeds**
* metadata with duplicate keyids works and keyid is double counted for threshold -> **Test Fails**

jku:
I think we can close this one: 
* test_duplicate_keys_root is very close to testing the _"fake reaching threshold with duplicate keyids"_ case
* #147 has been filed to make that test even more effective
* #143 has been filed for the inconsistent behaviour between clients WRT duplicate keyids

jku (changes requested):
marking "request changes" to remove my accidental approval: I think we might just be able to close this without merging

jku:
Closing, everything is handled by other issues

PR #109, opened 2024-08-05
Add linting

### LINKED ISSUE #6: Add linting
ruff+mypy has worked in other projects like python-tuf, that might be good here as well. Those projects use tox (https://github.com/theupdateframework/python-tuf/blob/main/tox.ini) but a Makefile should work as well

jku:
* Add ruff and mypy as optional dependencies
* Modify Makefile so it can install linters and has targets `lint` and `fix`
* Modify CI workflow so it runs `make lint`
* Mention linting in README
* Fix all issues that linters complain about
* Remove some dead code

This is only a start: the ruff default rulesets are not very strict but this initial commit already improves code quality so clearly this is useful.

Fixes #6 , should be ready for review now


tuf_conformance/metadata.py
```diff
@@ -28,7 +29,7 @@
 
 class MetadataTest(Metadata):
     @classmethod
-    def from_dict(cls, metadata: Dict[str, Any]) -> "MetadataTest[T]":
+    def from_dict(cls, metadata: Dict[str, Any]) -> "Metadata[T]":
```
AdamKorcz, line 32: What is the reason for this change?

tuf_conformance/metadata.py
```diff
@@ -109,7 +110,7 @@ def __init__(
 class JSONDeserializerTest(MetadataDeserializer):
     """Provides JSON to Metadata deserialize method."""
 
-    def deserialize(self, raw_data: bytes) -> MetadataTest:
+    def deserialize(self, raw_data: bytes) -> Metadata:
```
AdamKorcz, line 113: Same as https://github.com/theupdateframework/tuf-conformance/pull/109/files#r1704294902

tuf_conformance/metadata.py
jku: mypy complains about this type annotation when the method is called.. and I believe it is correct to complain: we do return a Metadata from the is method because we end up calling `Metadata()` at the end of the function.

jku: I could be wrong about the logic there, python inheritance has never been my strongest suit

jku: the error is 
> tuf_conformance/metadata.py:32: error: "MetadataTest" expects no type arguments, but 1 given  [type-arg]

which admittedly is a bit strange.

jku: hmmm... I think this might be related to the generic type, let me try a different tweak to fix this

jku: Yeah tweaking the inheritance annotation `class MetadataTest(Metadata[T])` makes mypy understand the intent as well. Fixed with a new commit.

AdamKorcz (approved):

PR #111, opened 2024-08-06
Refactor artifact download tests

jku:
This has multiple changes in the PR since they all require each other (not in individual commits, sorry about that):
* **example clients**: Do not default to no hash prefixes for target requests. The whole `prefix_targets_with_hash=False` configuration is technically a non-compliant feature added specifically for PyPI. We should not test for it: hash prefixes are an important part of TUF
* **utils**: Remove utility code to run additional web servers: per-test RepositorySimulator + a single SimulatorServer should give us everything we need (if they don't, let's improve those instead of adding new custom web servers)
* **ClientRunner**: target URL is delivered via ClientInitData in the same way that metadata URL already was: there is no need for tests to decide the target URL
* **target download tests**: Cleanup tests, refactor to use RepositorySimulator. A lot of lines change here but the test functionality should stay almost exactly the same


tuf_conformance/test_file_download.py
```diff
@@ -26,244 +13,70 @@ def test_client_downloads_expected_file(
...
-    # updating the file hashes and length
-    repo.targets.version += 1
-    repo.update_snapshot()
+    # Create legitimate test artifacts
+    target_path_1 = "target_file.txt"
+    target_content_1 = b"target file contents"
+    target_path_2 = "another_target_file.txt"
+    target_content_2 = b"content"
+    repo.add_target(Targets.type, target_content_1, target_path_1)
+    repo.add_target(Targets.type, target_content_2, target_path_2)
```
jku, line 50: this is a change compared to original test: we add two artifacts instead of one and initially client only downloads one of them.

This works around the coverage problem where in the original test "malicous" content was never downloaded (because the clients cache the downloads and the cached file was already the correct one). Now the second  file tests actually downloading malicious content and failing because the hashes do not match.

tuf_conformance/test_file_download.py
```diff
@@ -26,244 +13,70 @@ def test_client_downloads_expected_file(
...
+    with open(client.get_last_downloaded_target(), "rb") as last_download_file:
+        assert last_download_file.read() == target_content_1
 
-    # Now we check that the client has not downloaded the
-    # malicious file which is expected. We check the
-    with open(client.get_last_downloaded_target(), "r") as last_download_file:
-        assert last_download_file.read() == file_contents_str
-
-
-def test_repository_substitutes_target_file_and_bumps_10_times(
```
jku, line 179: ah, this test was actually removed: I don't think it was adding any real coverage.

AdamKorcz (approved):

Issue #106, opened 2024-08-05
decide file name encoding / path handling safety requirements


jku:
We expect client-under-test to write both metadata and artifacts onto disk for verification purposes (and we also expect most clients to do that as part of their normal operation). Unfortunately we don't specify the expectations in detail...

There are several decision to make that all relate to each other:
* what do we expect the `"download --target-name "path/file"` to do -- should it create the subdirectory "path"?
  * python-tuf as an example will not create the directory by default (instead it uses url encoded filename `path%2Ffile`)
* do we want to test url/filename encoding issues? For the record, I believe clients must URL encode rolenames and target paths since they are path of URL path fragments
  * what happens if a delegation uses rolenames like `"/"` `"?"` or `"😀"` ?
  * same question for artifacts (target paths)
* do we want to verify that client does not have path traversal security issues?
  * delegation name "../delegate" should not lead to writing in the higher level directory
  * target path "../filename" should not lead to writing in the higher level directory

As far as I know, the spec does not specify any of this. Testing these things feels important but I'm not sure how to do it without forcing clients into specific design choice.

**Possible design** 

I think at least for the role name cases we can avoid enforcing a specific filename encoding but 
* we can test that client never writes into the higher level directory even if we have a role named "../delegation"
* we can test some weird role names in a single test: this way clients can skip that test if they decide to not  support those

for artifacts I wonder if we should just change the CLI so that instead of `--target-dir` we have `--target-filename`: so the test suite specifically names the specific file name that it wants saved on disk?
* we can still test that strange target paths work (correct URL gets requested)
* but we avoid making a decision on how the file should be stored on disk by default (and as a result can't really test for path traversal issues in this case)

jku:
The related tests are not critical for 1.0 IMO but the CLI change decision should be made:
> I wonder if we should just change the CLI so that instead of `--target-dir` we have `--target-filename`: so the test suite specifically names the specific file name that it wants saved on disk?

jku:
**Alternative design**

Keep using --target-dir and lean more on the current design by @AdamKorcz:
* tests never expect a specific filename to be written on disk, instead they try to identify "last modified artifact"
* this "last modified artifact" lookup should work recursively unlike now (so clients can create subdirs if they wish to)
[...]

PR #113, opened 2024-08-06
Add download test using targetpath with subdirs

jku:
* modifies `get_last_downloaded_target()` so it's recursive, allowing clients to use subdirs if they want
* Now it also raises FileNotFound if nothing has been downloaded
* go-tuf seems to fail when trying to open() the file currently

Most of the download tests probably should be parametrized ( so at least the targetpath is parametrized) to minimzed copy-paste... but let's add a few tests first

This would be the "alternative design" in #106

### DIFF
--- a/clients/go-tuf/cmd/download.go
+++ b/clients/go-tuf/cmd/download.go
@@ -13,6 +13,7 @@ package cmd
 
 import (
 	"fmt"
+	"net/url"
 	"os"
 	"path/filepath"
 
@@ -103,7 +104,8 @@ func RefreshAndDownloadCmd(targetName string,
 	}
 
 	// target is available, so let's see if the target is already present locally
-	path, _, err := up.FindCachedTarget(targetInfo, "")
+	localPath := filepath.Join(targetDownloadDir, url.QueryEscape(targetName))
+	path, _, err := up.FindCachedTarget(targetInfo, localPath)
 	if err != nil {
 		return fmt.Errorf("failed while finding a cached target: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


jku:
It's unclear to me how go-tuf is supposed to be used:
* currently the client wrapper explicitly sets the file path in DownloadTarget() call: this path now contains sub directories and nothing creates the sub directories 
* filepath in DownloadTarget() call could be left as "", in that case the local path is generated within UpdaterConfig.LocalTargetsDir. However, setting UpdaterConfig.LocalTargetsDir to just about any value fails with `mkdir : no such file or directory` -- should the directory exist already?

jku:
I've added a workaround: I set the local path explicitly now in the go client wrapper (to a url-escaped filename so no subdirectories). 

I don't know if that's the right way to go about it, and something is clearly wrong (try fixing `cfg.LocalTargetsDir = FlagMetadataDir` and see), but this test now succeeds and I'm willing to call this approach to solving #106 possible

jku:
Ok, I think I'm done and I think this is doable:
* tests can now verify what content gets downloaded when
* for a simple case this is still simple: 
        assert client.get_downloaded_target_bytes() == [b"content"]
* for more complicated cases it's a bit harder but allows verifying that e.g. when artifact content was replaced by other content:
        expected_downloads.remove(b"old content")
        expected_downloads.append(b"new content")
        assert client.get_downloaded_target_bytes() == expected_downloads

AdamKorcz (approved):

PR #116, opened 2024-08-07
Rewrite the Client CLI documentation

jku:
* Aim to be useful to client implementers
* Remove internal details (ClientRunner etc): I don't object to documenting these better but I don't think the client CLI doc is the right place -- maybe in the code itself?
* Add examples of how the commands are called
* Add more details about the requirements for clients
* Move the CLI doc to root (to make it clearer that clients/ is an internal test suite thing)

The rendered version is in https://github.com/jku/tuf-conformance/blob/client-readme-update/CLIENT-CLI.md


CLIENT-CLI.md
```diff
@@ -0,0 +1,101 @@
...
+
+where
+- `METADATA_DIR` is a directory where the client is expected to store metadata it considers valid
+- `METADATA_URL` is a URL to the repository metadata store
+- `TARGET_PATH` is the TUF targetpath of the artifact that should be downloaded
+- `TARGET_URL` is the base URL for repository target store
+- `TARGET_DIR`: is a directory where the client should store downloaded and verified files
+
+Client must download the given artifact (targetpath) according to TUF client workflow, and store it in
+given targets directory. Client must ensure that metadata is up-to-date before downloading artifacts
```
jku, line 70: I'll highlight this: at least python-tuf should already operate like this but I think this might be the right choice overall:
* I think it more closely matches "normal" tuf client usage (just download the thing, don't make me jump through hoops)
* makes less assumptions about client implementations (do all clients even have separate refresh and download_target?)
* simplifies tests since there's fewer refresh() calls

jku:
@segiddins I can't mark you as a reviewer but feel free to have a look.

clients/README.md
```diff
@@ -1,50 +1,3 @@
...
-
-This project may then also run the test suite using the wrappers from the TUF implementations, but the intention is that this repository would not have to e.g. maintain dependency lists for the client wrappers.
-
-That said, some wrappers are included for now:
-* We will likely tweak the "wrapper API" while we build the initial test suite: while this goes on, it makes
-  sense to keep the wrappers here
-* We do need something to test the test suite itself
-* There is no easy way to run the test suite in a TUF implementation CI yet
\ No newline at end of file
+This directory contains test clients for the test suite: They are intended
```
AdamKorcz, line 1: Shall we add a link in `clients/README.md` to `CLIENT_CLI.md`?

jku: yeah, that makes sense

AdamKorcz (approved):

PR #122, opened 2024-08-10
Add more basic tests

jku:
1. This adds a bunch of basic tests -- something I think new clients should start with solving.
2. This also tries to cleanup the metadata access in RepositorySimulator:
   * avoid metadata access when the typed property is shorter and makes sense:  use `repo.timestamp.version` instead of `repo.md_timestamp.signed.version`
   * Store all metadata in a single dict: this way we don't need to have a five-way if-clause when we just need to access the role by name (which is the usual case in parametrized tests)
   * Try to provide easy accessors for typical cases (`repo.timestamp`, `repo.any_targets(role)`, etc)
   
I can separate the two if the review is annoying as is

### DIFF
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -402,9 +402,7 @@ class RepositorySimulator:
 
         # Add targets metadata for all bins.
         for delegated_name in succinct_roles.get_roles():
-            self.mds[delegated_name] = Metadata(
-                Targets(expires=self.safe_expiry)
-            )
+            self.mds[delegated_name] = Metadata(Targets(expires=self.safe_expiry))
 
             self.add_signer(delegated_name, signer)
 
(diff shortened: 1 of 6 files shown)


AdamKorcz (approved):

AdamKorcz:
Ah sorry, it is still in draft. Let me know when to review.

jku:
@AdamKorcz have a look if you want: I've rebased and added some docstrings, otherwise this is the same PR you already saw.

tuf_conformance/repository_simulator.py
```diff
@@ -108,25 +109,36 @@ def __init__(self, dump_dir: str | None) -> None:
 
     @property
     def root(self) -> Root:
-        return self.md_root.signed
+        signed = self.mds[Root.type].signed
+        assert isinstance(signed, Root)
+        return signed
```
jku, line 114: The asserts are there to keep type checker happy.

Issue #128, opened 2024-08-12
RepositorySimulator._compute_hashes_and_length is broken


jku:
EDIT: Every test that uses `RepositorySimulator._compute_hashes_and_length = True` is currently broken because :
* the simulator signs at request time
* ecdsa signatures are not deterministic
* hashes/length calculation is done over the whole file (so it includes these changing signatures)

This should get fixed if any option of #130 is implemented
 
Original report was :

---

test_snapshot_rollback_with_local_snapshot_hash_mismatch() does not seem right:
 * the metadata changes in the test seem valid, yet client fails refresh
 * test claims in comments to do a  target version rollback but does not?
 * client actually fails because length and hash are incorrect?
 
I can't quite see what the issue is. 

We should 
* find the root cause here
* fix the test 
* add a simple test that just uses hashes and length in metafiles without any fancy attacks

jku:
> find the root cause here

oh, it's likely the combination of:
* RepositorySimulator signing at request time
* using ecdsa keys that are not "deterministic"

the hashes/length  are over the whole file so include signatures.

jku:
I filed an issue for switching the signing keytype (or otherwise solving the underlying issue).

That said, the test is still completely wrong: it claims to test a targets rollback but does not actually try to do one.

AdamKorcz:
Should this remain open, @jku?

jku:
yeah, that makes sense: currently client fails as expected but for wrong reasons.

Once #131 is implemented (and #130 is fixed) we have more confidence that the failure happens for right reason

jku:
test_new_targets_hash_mismatch() is broken because of this as well: If any return values had been checked that would have been visible

Issue #130, opened 2024-08-13
Make signatures deterministic by default?


jku:
#128 seems to be an example of the flaw in current RepositorySimulator signing key & signature management:
* The design is optimized for minimal test code: this leads to signing metadata (apart from root) at request time
* The keys are created (in multiple places) with  `CryptoSigner.generate_ecdsa()` because ecdsa keys are fast to generate: this way each test simply generates whatever keys it needs

Unfortunately ecdsa signatures are not deterministic so the combination of these two features can easily lead to issues that are difficult to debug.

Obvious options to fix this include:
1. start explicitly signing metadata -- I would still like to avoid this if possible
2. switch to a deterministic signature by default

I think I lean to option 2, we should at least try this with another keytype:
* ed25519 is maybe not a great option, I'm not sure if every client has support for that. RSA is likely the right default
* the issue with RSA is the key generation time: in the long term we should not generate keys in every test
   * as the initial implementation we can try generating keys in tests like now, if it's not horribly slow
   * we could also include a list of ~20 hard coded keys in sources and have RepositorySimulator take keys from the list (keys are also generated in test_updater_key_rotations.py and test_keys.py: these should likely use the same source for key material)

jku:
> 1. start explicitly signing metadata

the upside of this is that we could more easily test e.g. #134

AdamKorcz:
> I would still like to avoid this if possible

Is this because of UX? - that it is simpler to write tests?

jku:
Yeah: I was expecting most tests to modify targets metadata (which is currently very easy), but I'm beginning to think the benefits may not be greater than the disadvantages... The fact that modifying root is different from other roles is already annoying.

I'd be happy to see a refactoring that does explicit signing in a clean way, but I have a feeling that's a large change -- at the very least it will touch every single test, and the repository http API will change slightly (as currently we return 404 on "old" version requests but if we have the old versions stored we would logically return them when asked). Would be nice to get the _hashes/length length in metafiles_ issue solved first (as that's currently actually broken and means our test results are partly incorrect).

We could do both options: first the second option (hard code RSA keys), and then look at a _explicit signing_ refactoring .

Opinions @AdamKorcz ?

jku:
> hard code RSA keys

alternative that has less hard coded key material would be to generate 20 signers _once_ in conftest.py and pass a copy of the list to all RepositorySimulators when they are created

jku:
[...]

PR #135, opened 2024-08-13
correct test_snapshot_rollback_with_local_snapshot_hash_mismatch so i…

AdamKorcz:
…t tests for a rollback

Fixes https://github.com/theupdateframework/tuf-conformance/issues/128#issuecomment-2285623486

### DIFF
--- a/tuf_conformance/test_rollback.py
+++ b/tuf_conformance/test_rollback.py
@@ -218,7 +218,7 @@ def test_snapshot_rollback_with_local_snapshot_hash_mismatch(
     repo.compute_metafile_hashes_length = True
 
     # The new targets must have a lower version than the local trusted one.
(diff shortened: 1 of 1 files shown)


jku (approved):
this is probably correct but we can't really verify that since the client will currently fail whatever you do for the wrong reason  ( because of #130  hashes in metafiles will always make a refresh fail)...

I guess we can merge blind?

AdamKorcz:
> this is probably correct but we can't really verify that since the client will currently fail whatever you do for the wrong reason ( because of https://github.com/theupdateframework/tuf-conformance/issues/130 hashes in metafiles will always make a refresh fail)...

Yes, until the hashes are fixed, it the client will fail because of that. 

> I guess we can merge blind?

I agree

Issue #132, opened 2024-08-13
test_new_snapshot_version_mismatch seems to do nothing


jku:
I don't understand what test_new_snapshot_version_mismatch() is trying to test for. I think it's incorrect but at least it's missing an explanation: It definitely has differences compared to the python-tuf test with same name (the original doesn't seem super useful either but it still looks reasonable)

PR #136, opened 2024-08-13
update comment to better explain intention of test

AdamKorcz:
Fix https://github.com/theupdateframework/tuf-conformance/issues/132

### DIFF
--- a/tuf_conformance/test_rollback.py
+++ b/tuf_conformance/test_rollback.py
@@ -159,6 +159,7 @@ def test_new_snapshot_version_mismatch(
 
     assert client.refresh(init_data) == 1
     assert client._files_exist([Root.type, Timestamp.type])
+    assert repo.metadata_statistics[-1] == (Snapshot.type, 1)
 
 
 def test_new_timestamp_fast_forward_recovery(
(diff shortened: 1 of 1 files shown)


jku (approved):
This looks sensible, happy to merge as is.

the `client._files_exist()` calls still feels misleading: the method does not actually ensure that the client only has these files. To be more explicit we could instead verify that client does not have trusted snapshot and check that `repo.metadata_statistics[-1] == (Snapshot.type, 1)`  (the client trying to download the correct snapshot version that our repo does not serve anymore)

AdamKorcz:
> To be more explicit we could instead verify that client does not have trusted snapshot and check that repo.metadata_statistics[-1] == (Snapshot.type, 1) (the client trying to download the correct snapshot version that our repo does not serve anymore)

I added that in stead of replacing. LMK WYT.

That being said, I believe we should have a client utility that checks that the client _only_ has the specified files.

jku:
> That being said, I believe we should have a client utility that checks that the client _only_ has the specified files.

I agree: it might make sense to replace _files_exist with a similar API as `metadata_statistics()` is on repository side: return list of tuples of rolename and version. Then tests could use to
* either check for existence of a specific role
* or compare the whole list to a list of expected trusted roles

Issue #137, opened 2024-08-13
reafctor ClientRunner._files_exist()


jku:
Copy pasting from #136

It might make sense to replace `ClientRunner._files_exist()` with a similar API as `metadata_statistics()` is on repository side: return list of tuples of rolename and version. Then tests could use to
* either check for existence of a specific role
* or compare the whole list to a list of expected trusted roles

latter is currently difficult.

jku:
I actually had a branch with something like this see #138

PR #138, opened 2024-08-13
Add unusual rolename tests

jku:
These tests are useful in codifying some expectations on how to handle weird rolenames. I wouldn't be surprised if some clients want to skip this test but the `"../delegatedrole"` case comes from an actual security issue we've had (file traversal when storing the metadata).

This also enables #137  although I did not replace usage of `_file_exist()` yet -- I'm not sure if the new  `trusted_roles(self)` is the right API: maybe it should be sorted by mtime instead of alpha?

### DIFF
--- a/tuf_conformance/client_runner.py
+++ b/tuf_conformance/client_runner.py
@@ -5,6 +5,7 @@ from collections.abc import Iterable
 from tempfile import TemporaryDirectory
 
 from tuf.api.exceptions import StorageError
+from tuf.api.metadata import Metadata
 
 from tuf_conformance.metadata import MetadataTest
 from tuf_conformance.simulator_server import ClientInitData, SimulatorServer
@@ -76,6 +77,7 @@ class ClientRunner:
         return self._run(cmd)
 
     def download_target(self, data: ClientInitData, target_name: str) -> int:
+        self._server.debug_dump(self.test_name)
         cmd = [
             *self._cmd,
             "--metadata-url",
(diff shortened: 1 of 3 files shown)


tuf_conformance/client_runner.py
```diff
@@ -110,7 +112,16 @@ def _files_exist(self, roles: Iterable[str]) -> bool:
         local_metadata_files = sorted(os.listdir(self.metadata_dir))
         return all(x in local_metadata_files for x in expected_files)
 
-    def _content(self, role: str) -> bytes:
-        """Return role metadata as bytes"""
-        with open(os.path.join(self.metadata_dir, f"{role}.json"), "rb") as f:
-            return f.read()
+    def trusted_roles(self) -> dict[str, int]:
```
jku, line 115: I suppose this return value  could be `list[tuple(str, int)]` and the list could be sorted by mtime to match what `RepositorySimulator.metadata_statistics() `returns

tuf_conformance/test_quoting_issues.py
```diff
@@ -0,0 +1,48 @@
...
+
+import pytest
+from tuf.api.metadata import DelegatedRole, Targets
+
+from tuf_conformance.client_runner import ClientRunner
+from tuf_conformance.simulator_server import SimulatorServer
+
+# tuple of
+#  * name of delegated role,
+#  * do we expect the role to be found in client trusted roles
```
jku, line 11: This was left over from a previous version: it's incorrect

PR #139, opened 2024-08-13
Add target search test

AdamKorcz:
Migrates [this test](https://github.com/theupdateframework/python-tuf/blob/cc153d71dc607b681b1c2803f21eb58f4526007f/tests/test_updater_delegation_graphs.py#L511) from python-tuf.

Depends on https://github.com/theupdateframework/tuf-conformance/pull/115 to be merged first.

PR #140, opened 2024-08-13
Add target search test

AdamKorcz:
Migrates [this test](https://github.com/theupdateframework/python-tuf/blob/cc153d71dc607b681b1c2803f21eb58f4526007f/tests/test_updater_delegation_graphs.py#L511) from python-tuf.

Replaces https://github.com/theupdateframework/tuf-conformance/pull/139

With this PR, the client stores the target_info locally in a file and overwrites it each time. This works for this test, but it is not a universal solution.

### DIFF
--- a/tuf_conformance/test_updater_delegation_graphs.py
+++ b/tuf_conformance/test_updater_delegation_graphs.py
@@ -4,15 +4,13 @@ import pytest
 from tuf.api.metadata import (
     SPECIFICATION_VERSION,
     DelegatedRole,
-    TOP_LEVEL_ROLE_NAMES,
-    DelegatedRole,
-    TargetFile,
     Targets,
 )
 
 from tuf_conformance.client_runner import ClientRunner
 from tuf_conformance.repository_simulator import RepositorySimulator
 from tuf_conformance.simulator_server import SimulatorServer
+
 
 @dataclass
 class DelegationTester:
@@ -50,6 +48,7 @@ class DelegationsTestCase:
 
 # DataSet is only here so type hints can be used.
 DataSet = dict[str, DelegationsTestCase]
+DataSetTarget = dict[str, TargetTestCase]
 
 
 graphs: DataSet = {
@@ -225,15 +224,15 @@ delegations_tree = DelegationsTestCase(
         DelegationTester("B", "D", paths=["releases/y/*.zip"]),
     ],
     target_files=[
-        TestTarget("targets", b"targetfile content", "targetfile"),
-        TestTarget("A", b"README by A", "README.md"),
-        TestTarget("C", b"x release by C", "releases/x/x_v1"),
-        TestTarget("D", b"y release by D", "releases/y/y_v1.zip"),
-        TestTarget("D", b"z release by D", "releases/z/z_v1.zip"),
+        TargetTest("targets", b"targetfile content", "targetfile"),
+        TargetTest("A", b"README by A", "README.md"),
+        TargetTest("C", b"x release by C", "releases/x/x_v1"),
+        TargetTest("D", b"y release by D", "releases/y/y_v1.zip"),
+        TargetTest("D", b"z release by D", "releases/z/z_v1.zip"),
     ],
 )
 
-targets: DataSet = {
+targets: DataSetTarget = {
     "no delegations": TargetTestCase("targetfile", True, []),
     "targetpath matches wildcard": TargetTestCase("README.md", True, ["A"]),
     "targetpath with separators x": TargetTestCase("releases/x/x_v1", True, ["B", "C"]),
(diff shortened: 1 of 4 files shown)


jku (changes requested):
Just like in a previous PR, I do not think it is acceptable to require targetinfo to be available: this is a change to the client CLI and we do not want those.  

I also don't see why we absolutely need targetinfo: in python-tuf it is possible to make tests more "unit testy" so there testing just the artifact search is a good idea. This is a conformance test suite that should do end-to-end tests: in this case it means instead of "artifact search" we  test "artifact download".

Let's continue with this after #115 is merged (all the same issues will apply to this)

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+        local_target_file = TargetFile.from_data(
+            downloaded_target_path, downloaded_target_bytes
+        )
+        assert local_target_file.to_dict() == exp_target.to_dict()
+    else:
+        assert client.download_target(init_data, target.targetpath) == 1
+
+    # repo prepends and appends [('root', 2), ('timestamp', None)]
+    # so we compare equality from the 2nd call to the 3rd-last
+    # call.
```
jku, line 149: this is just the client making sure metadata is fresh on two different download_target calls

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+    assert repo.metadata_statistics[2:-2] == exp_calls
+    assert client._files_exist(exp_files)
```
jku, line 151: should rebase and use `trusted_roles()`

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+    init_repo(repo, delegations_tree)
+    exp_target = repo.artifacts[target.targetpath].target_file
+
+    # Call explicitly refresh to simplify the expected_calls list
+    client.refresh(init_data)
+    repo.metadata_statistics.clear()
+    client.download_target(init_data, target.targetpath)
+    if target.found:
+        assert client.download_target(init_data, target.targetpath) == 0
+        downloaded_target_path = client.get_downloaded_target_path()
```
jku, line 138: I still don't think we should do this. We decided in #106 that 

> tests never expect a specific filename to be written on disk

This is mostly because some clients may create a subdirectory structure for the targetpath, some might not (but also because of encoding issues in filenames). So we should not compare the filename to anything

Comparing the bytes should be enough, right?

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+        downloaded_target_bytes = client.get_downloaded_target_bytes()[0]
+        local_target_file = TargetFile.from_data(
+            downloaded_target_path, downloaded_target_bytes
+        )
+        assert local_target_file.to_dict() == exp_target.to_dict()
```
jku, line 143: would this work?
```suggestion
        assert client.get_downloaded_target_bytes() == [
           repo.artifacts[target.targetpath].data
        ]
```

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+            targets = Targets(1, spec_version, repo.safe_expiry, {}, None)
+        # unpack 'd' but skip "delegator"
+        role = DelegatedRole(*astuple(d)[1:])
+        repo.add_delegation(d.delegator, role, targets)
+
+    for target in test_case.target_files:
+        repo.add_artifact(*astuple(target))
+
+    if test_case.target_files:
+        repo.targets.version += 1
```
jku, line 71: I think this is unnecessary for reasons mentioned in the other PR

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+    rolename: str
+    keyids: list[str] = field(default_factory=list)
+    threshold: int = 1
+    terminating: bool = False
+    paths: list[str] | None = field(default_factory=lambda: ["*"])
+    path_hash_prefixes: list[str] | None = None
+
+
+@dataclass
+class TestTarget:
```
jku, line 33:
```
test_updater_delegation_graphs.py:32: PytestCollectionWarning: cannot collect test class 'TestTarget' because it has a __init__ constructor (from: tuf_conformance/test_updater_delegation_graphs.py)
    @dataclass
```

jku (changes requested):

tuf_conformance/test_updater_delegation_graphs.py
jku: could also assert about repo.artifact_statistics if that's not too complicated

AdamKorcz: Removed the file check

tuf_conformance/test_updater_delegation_graphs.py
AdamKorcz: remove `and appends`

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,151 @@
...
+
+    init_data, repo = server.new_test(client.test_name)
+    assert client.init_client(init_data) == 0
+    init_repo(repo, delegations_tree)
+    exp_target = repo.artifacts[target.targetpath].target_file
+
+    # Call explicitly refresh to simplify the expected_calls list
+    client.refresh(init_data)
+    repo.metadata_statistics.clear()
+    client.download_target(init_data, target.targetpath)
```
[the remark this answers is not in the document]
AdamKorcz: Removed that call.

tuf_conformance/test_updater_delegation_graphs.py
AdamKorcz: Added https://github.com/theupdateframework/tuf-conformance/pull/140#discussion_r1717372833 instead.

tuf_conformance/test_updater_delegation_graphs.py
AdamKorcz: Yes, thank you! Added

tuf_conformance/test_updater_delegation_graphs.py
AdamKorcz: Renamed to `TargetTest`

tuf_conformance/test_updater_delegation_graphs.py
AdamKorcz: Removed in previous PR. Still not required here.

jku (approved):
Test code looks good to me.

We could have quite a few more tests in the data set but this seems ok as a start.

Issue #127, opened 2024-08-12
Test succinct delegations


jku:
This is not on any critical path (as I don't know of repositories using succinct delegations) but:

We should add a basic test for succinct delegations at some point:
* I don't remember how the spec defines this but I would consider this an optional feature of the spec so should make this easy to make "expected to fail"
* there is some functionality in RepositorySimulator to create these things already

jku:
> I don't remember how the spec defines this but I would consider this an optional feature of the spec so should make this easy to make "expected to fail"

Turns out this is still an augmentation proposal: we'll have to figure out how we handle optional features.
My first instinct is just action inputs like
```
tap15-succinct-hashed-bin-delegations:
   default: "false"
```

jku:
I wanted to close the PR since it's unlikely to move forward before we have a solution for the "optional features" but I also wanted to preserve the work: it's now in [succint-graph-traversal](https://github.com/theupdateframework/tuf-conformance/tree/succint-graph-traversal) branch.

jku:
> it's unlikely to move forward before we have a solution for the "optional features" 

It looks like there's not a lot of demand for "optional features" flags so I guess we can handle this one with just the existing "xfails" approach: clients that don't support succinct delegations can just mark the test as expected to fail.

So the existing branch could be rescued: 
* it will need fixes because infra code has changed a bit
* there are comments in #141 that would need to be addressed

PR #141, opened 2024-08-13
add test for succint role test case

AdamKorcz:
Migrates this test from python-tufs test suite: https://github.com/theupdateframework/python-tuf/blob/cc153d71dc607b681b1c2803f21eb58f4526007f/tests/test_updater_delegation_graphs.py#L473

Work on https://github.com/theupdateframework/tuf-conformance/issues/127

### DIFF
--- a/tuf_conformance/test_updater_delegation_graphs.py
+++ b/tuf_conformance/test_updater_delegation_graphs.py
@@ -1,0 +1,127 @@
+from dataclasses import dataclass, field
+from typing import Any
+
+import pytest
+from tuf.api.metadata import (
+    TOP_LEVEL_ROLE_NAMES,
+)
+
+from tuf_conformance.client_runner import ClientRunner
+from tuf_conformance.simulator_server import SimulatorServer
+
+# DataSet is only here so type hints can be used.
+DataSet = dict[str, Any]
+
+
+@dataclass
+class DelegationTester:
+    delegator: str
+    rolename: str
+    keyids: list[str] = field(default_factory=list)
+    threshold: int = 1
+    terminating: bool = False
+    paths: list[str] | None = field(default_factory=lambda: ["*"])
+    path_hash_prefixes: list[str] | None = field(default_factory=list)
+
+
+@dataclass
+class TestTarget:
+    rolename: str
+    content: bytes
+    targetpath: str
+
+
+@dataclass
+class SuccinctRolesTestCase:
+    bit_length: int
+    target_path: str
+    expected_target_bin: str
+
+
+@dataclass
+class DelegationsTestCase:
+    """A delegations graph as lists of delegations and target files
+    and the expected order of traversal as a list of role names."""
+
+    delegations: list[DelegationTester]
+    target_files: list[TestTarget] = field(default_factory=list)
+    visited_order: list[str] = field(default_factory=list)
+
+
+succinct_bins_graph: DataSet = {
+    "bin amount = 2, taget bin index 0": SuccinctRolesTestCase(
+        bit_length=1,
+        target_path="boo",
+        expected_target_bin="bin-0",
+    ),
+    "bin amount = 2, taget bin index 1": SuccinctRolesTestCase(
+        bit_length=1,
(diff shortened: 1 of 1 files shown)


tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,127 @@
...
+    assert client.refresh(init_data) == 0
+    repo.metadata_statistics.clear()
+    # Check that metadata dir contains only top-level roles
+    assert client._files_exist(TOP_LEVEL_ROLE_NAMES)
+
+    # Looking for a non-existing targetpath forces updater
+    # to visit a corresponding delegated role.
+    client.download_target(init_data, succinct_bins_graph.target_path)
+    # Here we could check that target_info is null. TODO: Add when
+    # support for storing the target_info is added.
```
jku, line 122: I would check the return value.

I am against support for storing targetinfo

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,127 @@
...
+    repo.add_succinct_roles("targets", succinct_bins_graph.bit_length, "bin")
+    repo.update_snapshot()
+    assert client.init_client(init_data) == 0
+    # Call explicitly refresh to simplify the expected_calls list.
+    assert client.refresh(init_data) == 0
+    repo.metadata_statistics.clear()
+    # Check that metadata dir contains only top-level roles
+    assert client._files_exist(TOP_LEVEL_ROLE_NAMES)
+
+    # Looking for a non-existing targetpath forces updater
```
jku, line 118: nit: "updater" is a python-tuf component

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,127 @@
...
+    # to visit a corresponding delegated role.
+    client.download_target(init_data, succinct_bins_graph.target_path)
+    # Here we could check that target_info is null. TODO: Add when
+    # support for storing the target_info is added.
+
+    # Check that the delegated roles were visited in the expected
+    # order and the corresponding metadata files were persisted.
+    assert repo.metadata_statistics[2:] == exp_calls
+    assert client._files_exist(exp_files)
```
jku, line 127: `_files_exist()` does not verify that only these files exist so I think it adds little value here. I would remove it or wait for #138 where I add a ClientRunner method for that

tuf_conformance/test_updater_delegation_graphs.py
```diff
@@ -0,0 +1,127 @@
...
+    rolename: str
+    keyids: list[str] = field(default_factory=list)
+    threshold: int = 1
+    terminating: bool = False
+    paths: list[str] | None = field(default_factory=lambda: ["*"])
+    path_hash_prefixes: list[str] | None = field(default_factory=list)
+
+
+@dataclass
+class TestTarget:
```
jku, line 28: does this name not make pytest warn?

jku (commented):
Looks about right, left some minor comments.

However, I realized only now that the succinct delegations have not made it to the spec yet. I do want this in the test suite  but we likely can't have it on by default, that would be annoying to clients who don't implement it.

Sorry to leave another PR hanging but we'll have to figure out how deal with optional features: maybe we add action inputs for them...

tuf_conformance/test_updater_delegation_graphs.py
jku:
```
PytestCollectionWarning: cannot collect test class 'TestTarget' because it has a __init__ constructor (from: tuf_conformance/test_updater_delegation_graphs.py)
```

jku (changes requested):
I'll mark this "changes requested":
* We need a solution for "optional features" before we add this
* optional features must be togglable in the action and in the command line
* default is likely "don't enable"

Since this is not part of the core spec, it's also not that important to work on

jku:
I'm going to close this to tidy up: the work so far is preserved in succint-graph-traversal branch and the issue remains open

PR #149, opened 2024-08-14
Specify installation of python dependencies from conformance test pyproject.toml

segiddins:
Fixes usage when project under test has no python deps specification


segiddins:
Does not work as intended :/

PR #150, opened 2024-08-15
Static tests

jku:
* Add some infrastructure support for testing static repositories: I may have gone overkill but I made the same kind of setup as the simulated repositories have (so a client fixture and server fixture). ClientRunner works with both.
* The end result is that the test itself requires no changes when new static data is added: Just add the repository content into a sub directory of `tuf_conformance/static_data` and it will be treated as a new test
* Add test data for tuf-on-ci: the idea is that the metadata is mostly fire-and-forget -- once it's committed it stays there or we just wipe it and replace with something new
* The tested surface could be increased: currently there is a single targetpath that is expected to successfully download 

@kairoaraujo I'm planning this sort of setup for some practical compatibility tests -- so repository software developers (or specific repository maintainers) can provide example data that clients can test against. Let me know if you have comments

### DIFF
--- a/tuf_conformance/test_static_repositories.py
+++ b/tuf_conformance/test_static_repositories.py
@@ -1,0 +1,22 @@
+import os
+
+import pytest
+
+from tuf_conformance.client_runner import ClientRunner
+from tuf_conformance.simulator_server import StaticServer
+
+static_repos = []
+for static_dir in os.listdir(os.path.join("tuf_conformance", "static_data")):
+    if os.path.isdir(os.path.join("tuf_conformance", "static_data", static_dir)):
+        static_repos.append(static_dir)
+
+
+@pytest.mark.parametrize("static_repo", static_repos)
+def test_static_repo(
+    static_client: ClientRunner, static_server: StaticServer, static_repo: str
+) -> None:
+    init_data, targetpath = static_server.new_test(static_repo)
+
+    assert static_client.init_client(init_data) == 0
+    assert static_client.refresh(init_data) == 0
+    assert static_client.download_target(init_data, targetpath) == 0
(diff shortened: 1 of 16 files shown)


kairoaraujo:
Hi @jku,
I can add some static test data from RSTUF as well.
I'll find some time this weekend.

jku:
> I can add some static test data from RSTUF as well. I'll find some time this weekend.

Sure, no rush -- just wanted to ping in case you see anything in the design that might not work for you

AdamKorcz (approved):
Nice!

tuf_conformance/test_static_repositories.py
```diff
@@ -0,0 +1,27 @@
+import os
+
+import pytest
+
+from tuf_conformance.client_runner import ClientRunner
+from tuf_conformance.simulator_server import StaticServer
+
+static_repos = []
+for static_dir in os.listdir(os.path.join("tuf_conformance", "static_data")):
```
segiddins, line 9: I dont think this works when running the action from another repo, see: https://github.com/sigstore/sigstore-ruby/actions/runs/10457763755/job/28958137257?pr=69

```
_________ ERROR collecting tuf_conformance/test_static_repositories.py _________
../../_actions/theupdateframework/tuf-conformance/07b2952eda3f971c9bade669486643c28bf40a1d/tuf_conformance/test_static_repositories.py:9: in <module>
    for static_dir in os.listdir(os.path.join("tuf_conformance", "static_data")):
E   FileNotFoundError: [Errno 2] No such file or directory: 'tuf_conformance/static_data'
```

jku: thanks that seems obvious now that you point it out: #156

PR #151, opened 2024-08-16
Fix and extend targets rollback test

jku:
* Fix the test for targets rollback when metafile hashes are used
* Add a parametrization for targets and snapshot rollback tests: they now test both without hashes in metafiles and with hashes in metafiles (the test is 100% the same, only difference is whether repsoitory uses hashes)

This now leads to the "with hashes" tests failing because of #128: I believe the tests are correct.  This PR can be used to test the fix for #128. 

This PR is ready for review but should not be merged before the fix for #128.


jku:
closing for #154

PR #154, opened 2024-08-19
Change default keytype to RSA

jku:
This is to get deterministic signatures: this is a workaround to get the hash tests fixed now (#128). The change of default might be temporary but all other changes are still useful if we switch back to default ecdsa later (like only generating 8 keys instead of 600 keys):
    * Use RSA as the default keytype
    * Create a list of RSA signers at repository_simulator.py import time: Use these in all tests to avoid generating keys in every test
    * Fix the hash test in test_basic.py: it was incorrect.
    * Fix and parametrize the rollback tests (these are the commits from #151)
    
The last bullet point verifies that this works as I wrote the tests in #151 so they fail with main: now they succeed with this keytype change

### DIFF
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -151,7 +151,7 @@ class RepositorySimulator:
         self.root.roles[role].keyids.clear()
         self.signers[role].clear()
         for _ in range(0, self.root.roles[role].threshold):
-            signer = CryptoSigner.generate_ecdsa()
+            signer = CryptoSigner.generate_rsa(scheme="rsa-pkcs1v15-sha256")
             self.root.add_key(signer.public_key, role)
             self.add_signer(role, signer)
 
@@ -164,7 +164,7 @@ class RepositorySimulator:
         self.mds[Root.type] = MetadataTest(RootTest(expires=self.safe_expiry))
 
         for role in TOP_LEVEL_ROLE_NAMES:
-            signer = CryptoSigner.generate_ecdsa()
+            signer = CryptoSigner.generate_rsa(scheme="rsa-pkcs1v15-sha256")
             self.root.add_key(signer.public_key, role)
             self.add_signer(role, signer)
 
@@ -338,7 +338,7 @@ class RepositorySimulator:
         delegator.delegations.roles[role.name] = role
 
         # By default add one new key for the role
-        signer = CryptoSigner.generate_ecdsa()
+        signer = CryptoSigner.generate_rsa(scheme="rsa-pkcs1v15-sha256")
         delegator.add_key(signer.public_key, role.name)
         self.add_signer(role.name, signer)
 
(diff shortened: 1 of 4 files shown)


jku:
I'm leaving this a draft until we know if this is acceptable to known external users...

jku:
Hmm, this is missing a commit or two, investigating...

tuf_conformance/test_basic.py
```diff
@@ -15,7 +15,7 @@ def test_basic_init_and_refresh(client: ClientRunner, server: SimulatorServer) -
     Run a refresh, verify client trusted metadata and requests made by the client
     """
     init_data, repo = server.new_test(client.test_name)
-
+    print(init_data.trusted_root)
```
segiddins, line 18: debugging print

tuf_conformance/repository_simulator.py
```diff
@@ -151,7 +158,7 @@ def rotate_keys(self, role: str) -> None:
         self.root.roles[role].keyids.clear()
         self.signers[role].clear()
         for _ in range(0, self.root.roles[role].threshold):
-            signer = CryptoSigner.generate_ecdsa()
+            signer = self._rsa_signers.pop()
```
segiddins, line 161: maybe wrap `_rsa_signers.pop()` in a `checkout_key()` method, so it becomes easier to parameterize key types? & also to error in a helpful way if more than N keys are needed

segiddins (approved):

segiddins (approved):

jku:
last force push fixes a naming mixup in exception message

AdamKorcz (approved):

jku:
I just found out there is a PR to remove support for this keytype in go-tuf: This seems a bit reckless to me but let's make sure that's not the plan ...

jku:
No, reading the PR code, it does keep the compatibility. We are good to go

tuf_conformance/test_basic.py
```diff
@@ -166,29 +165,29 @@ def test_timestamp_content_changes(
...
     repo.snapshot.meta["targets.json"].version = repo.targets.version
-    repo.snapshot.version += 1
+    repo.snapshot.version += 1  # v3
     repo.update_timestamp()
 
-    client.refresh(init_data)
-    assert client.version(Snapshot.type) == 1
+    # Verify client refuses targets that does not match hashes
+    assert client.refresh(init_data) == 1
+    assert client.version(Snapshot.type) == 3
```
segiddins, line 190: shouldn't this be v2, since the refresh fails due to the snapshot hash mismatch?

jku: The mismatch is in targets hash (that is stored in snapshot) -- the docstring could make this clearer:
* snapshot v3 is a valid snapshot (with meta that contains targets v2).
* Updating to targets v2 fails because it does not match the hash/length in snapshot meta.

segiddins: ah i think this is actually surfacing a bug in my implementation!

segiddins: https://github.com/sigstore/sigstore-ruby/pull/89

PR #158, opened 2024-08-20
add assertions to client refreshes

AdamKorcz:
With this PR, all client refreshes should have assertions.

### DIFF
--- a/tuf_conformance/test_basic.py
+++ b/tuf_conformance/test_basic.py
@@ -153,14 +153,14 @@ def test_timestamp_content_changes(
     init_data, repo = server.new_test(client.test_name)
 
     assert client.init_client(init_data) == 0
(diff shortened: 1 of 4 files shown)


tuf_conformance/test_basic.py
```diff
@@ -187,7 +187,7 @@ def test_new_targets_hash_mismatch(
     repo.snapshot.version += 1
     repo.update_timestamp()
 
-    client.refresh(init_data)
+    assert client.refresh(init_data) == 1
```
jku: same issue as previous

jku (changes requested):
looks fine otherwise but the asserts for hash using tests are incorrect. Rebasing after #128 fix should help

jku (approved):

PR #163, opened 2024-08-21
Rewrite test docstring

AdamKorcz:
Attempt to make the intent of the test clearer by:
- Avoiding the mention of "it" which can be misleading. I wasn't sure what it referred to the first time I read the comment.
- Use "Ensures" instead of "Tests" and "Checks" "... whether". This should make it more clear what the correct behaviour is.

We could also use "higher version" instead of "newer version".


jku (approved):
thanks

Issue #33, opened 2024-07-03
tests: Test optional features


jku:
Specification has a few optional features. This list is from memory so possibly not complete:

**hashes & length in METAFILEs**

The single-item dictionary in timestamp and the dictionary in snapshot may or may not contain hashes and/or length. We should check that at least both reasonable options are supported (neither hashes and nor length; and both hashes and length included)

**Consistent snapshot**

According to spec the "consistent snapshot" feature is SHOULD only. I don't think we're very interested in the "non-consistent" version: I think we should only possibly test it with security in mind -- but have no practical examples at the moment

jku:
hashes/length: #131

jku:
hash algorithms is another unspecified  thing (spec does not even recommend what to use): #170

jku:
See also #141

jku:
Updating current status: `xfails` approach seems work just fine for individual tests covering two cases
  * actual conformance tests that client is currently failing
  * tests like `test_artifact_cache` that re not strictly conformance but features client can decide not to support

The remaining question is whether new flags are needed for larger optional features (that would enable/disable multiple tests at a time)
  * succinct delegations
  * non-consistent snapshot support
  * ?

I'm leaning towards not adding any feature flags now: non-consistent snapshot is something I want to discourage and not promote, so the only remaining use case would be succinct delegations: I'm sure clients can add those tests into xfails without trouble when they are added

PR #178, opened 2024-08-27
Improve the "expected failures" experience

jku:
If we add more tests that we consider more or less optional features, we need a way for clients to manage them."xfails" seems like it should work for that: Make it possible to write the expected failures to a file next to the client-under-test executable
* This should make managing a longer xfail list reasonable.
* This also means we don't need to repeat the xfails for go-tuf in two places.

I'm happy to get other opinions on this too: I have no idea if there are best practices for this sort of thing, I just made this up...

---
This is a change in the GitHub Action API (since I removed the `expected-failures` input) so we should bump major version. That said, I don't think any outside user was using the input yet.

Should make #33 more feasible

### DIFF
--- a/tuf_conformance/conftest.py
+++ b/tuf_conformance/conftest.py
@@ -17,13 +17,6 @@ def pytest_addoption(parser: pytest.Parser) -> None:
         type=str,
     )
     parser.addoption(
-        "--expected-failures",
-        action="store",
-        help="Optional space delimited list of test names expected to fail",
-        required=False,
-        type=str,
-    )
-    parser.addoption(
         "--repository-dump-dir",
         action="store",
         help="Optional path to dump repository versions for each test for debugging",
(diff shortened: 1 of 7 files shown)


tuf_conformance/conftest.py
```diff
@@ -83,14 +77,25 @@ def static_client(
...
+@cache
+def read_xfails(pytestconfig: pytest.Config) -> list[str]:
+    # Find expected failures from .xfails file
+    xfail_file = f"{pytestconfig.getoption('--entrypoint')}.xfails"
+    if not os.path.isabs(xfail_file):
+        xfail_file = os.path.join(pytestconfig.invocation_params.dir, xfail_file)
+
+    try:
+        with open(xfail_file) as f:
+            return f.read().splitlines()
```
jku, line 89: it might be useful to parse the content a bit -- e.g. allow comments so you could do something like this

```
# RSA PSS not yet supported https://github.com/theupdateframework/go-tuf/pull/625 
test_keytype_and_scheme[rsa/rsassa-pss-sha256]
```
in the .xfails file... but maybe we add that when someone actually wants it

jku:
rebased on main

AdamKorcz (approved):

PR #188, opened 2024-09-04
switch to explicit publishing

AdamKorcz:
Adds a `signed_mds` value to the `RepositorySimulator` which contains the repositorys released metadata. With this PR, the repository will make changes locally, and the `publish` releases the current state. Prior to this PR, `RepositorySimulator` publishes/signs the metadata when the client requests, but for some tests this has been to much of a constraint. 

`publish` takes a list of the roles it should publish, and it can also bump the roles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,14 +83,18 @@ def test_targets_version(
     # Initialize our repository: modify targets version and make sure the version is included in snapshot
     init_data, repo = server.new_test(client.test_name)
     repo.targets.version = 7
-    repo.update_snapshot()  # snapshot v2
+    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
 
     # initialize client-under-test, run refresh
     client.init_client(init_data)
     client.refresh(init_data)
 ```
 
-we can now run the test:
+The test initializes the repo and the initial data. It then sets the repo targets version to 7 and publishes the Targets, Snapshot and Timestamp roles. Finally, it initializes the client and refreshes the clients local metadata.
+`repo.publish()` makes the repositorys data public to the client, and it bumps the version of the metadata in the list by one. When a test needs to publish and make the changes available to the client, the `repo.publish()` call must include the role it wants to publish as the first item in the list followed by the roles the client updates before it ordered by last to first. I.e. in order to publish the Targets role, we must also publish Snapshot and Timestamp in that order. `repo.publish()` can also publish the Root role.
+
+We can now run the test:
+
 ```bash
 ./env/bin/pytest tuf_conformance \
     -k test_targets_version                 # run a specific test only
@@ -102,12 +106,12 @@ cat /tmp/test-repos/test_targets_version/refresh-1/targets.json
 cat /tmp/test-repos/test_targets_version/refresh-1/snapshot.json
 ```
 
-The metadata looks as expected (targets version is 7) so we can add a modification to the end of the test:
+The metadata looks as expected (targets version is 8 because we set the version to 7 and then bumped when we invoked `repo.publish()`) so we can add a modification to the end of the test:
 
 ```python
     # Make an non-compliant change in repository
     repo.targets.version = 6
-    repo.update_snapshot() # snapshot v3
+    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
 
     # refresh client again
     client.refresh(init_data)
@@ -135,7 +139,7 @@ def test_targets_version(
     # Initialize our repository: modify targets version and make sure it's included in snapshot
     init_data, repo = server.new_test(client.test_name)
     repo.targets.version = 7
-    repo.update_snapshot()  # snapshot v2
+    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
 
     # initialize client-under-test
     client.init_client(init_data)
@@ -145,12 +149,12 @@ def test_targets_version(
 
     # Make a non-compliant change in repository
     repo.targets.version = 6
-    repo.update_snapshot() # snapshot v3
+    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
 
     # refresh client again, expect failure and refusal to accept snapshot and targets
     assert client.refresh(init_data) == 1
-    assert client.version(Snapshot.type) == 2
-    assert client.version(Targets.type) == 7
+    assert client.version(Snapshot.type) == 3
+    assert client.version(Targets.type) == 8
 ```
 
 ### Releasing
--- a/README.md
+++ b/README.md
@@ -73,25 +73,20 @@ make fix
 
 ### Creating (and debugging) tests
 
-Let's say we want to test that clients will not accept a decreasing targets version. We start with a simple skeleton
-test that doesn't assert anything but sets up a repository and runs client refresh against it:
+Let's say we want to test that clients will not accept a decreasing targets version. We start with a simple skeleton test that sets up a repository and runs client refresh against it:
 
 ```python
-def test_targets_version(
+def test_targets_version_rollback(
     client: ClientRunner, server: SimulatorServer
 ) -> None:
-    # Initialize our repository: modify targets version and make sure the version is included in snapshot
+    # Initialize our repository: modify targets version and make sure it's included in snapshot
     init_data, repo = server.new_test(client.test_name)
-    repo.targets.version = 7
-    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
 
-    # initialize client-under-test, run refresh
+    # initialize client-under-test
     client.init_client(init_data)
-    client.refresh(init_data)
+    # Run refresh on client-under-test, expect success
+    assert client.refresh(init_data) == 0
 ```
-
-The test initializes the repo and the initial data. It then sets the repo targets version to 7 and publishes the Targets, Snapshot and Timestamp roles. Finally, it initializes the client and refreshes the clients local metadata.
-`repo.publish()` makes the repositorys data public to the client, and it bumps the version of the metadata in the list by one. When a test needs to publish and make the changes available to the client, the `repo.publish()` call must include the role it wants to publish as the first item in the list followed by the roles the client updates before it ordered by last to first. I.e. in order to publish the Targets role, we must also publish Snapshot and Timestamp in that order. `repo.publish()` can also publish the Root role.
 
 We can now run the test:
 
@@ -106,15 +101,33 @@ cat /tmp/test-repos/test_targets_version/refresh-1/targets.json
 cat /tmp/test-repos/test_targets_version/refresh-1/snapshot.json
 ```
 
-The metadata looks as expected (targets version is 8 because we set the version to 7 and then bumped when we invoked `repo.publish()`) so we can add a modification to the end of the test:
+Now, the repository publishes Targets 6 times so that it is version 7. The client then refreshes and the test asserts that the client has Targets version 7. Next, the repository attempts a rollback attack; It attempts to make the client download metadata for the Targets role that is a lower version than the version the client already has. To do that, the test deletes the already published Targets and republishes 5 times. Now, the repositorys Targets version is 5. The test publishes Snapshot and Timestamp too so the client can see the latest Targets changes. The client now refreshes which must fail to make the client spec-compliant. Finally, the test asserts that the client still has version 7. 
 
 ```python
-    # Make an non-compliant change in repository
-    repo.targets.version = 6
-    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
+  # Bump version of Targest 6 times.
+    for i in range(6):
+        repo.publish([Targets.type])
+    repo.publish([Snapshot.type, Timestamp.type])
+    
+    # Client refreshes and downloads Targets version 7
+    assert client.refresh(init_data) == 0
+    assert client.version(Targets.type) == 7
 
-    # refresh client again
-    client.refresh(init_data)
+    # The repository deletes all published Targets
+    # and republishes to version 5.
+    # This is a rollback attack.
+    del repo.signed_mds[Targets.type]
+    for i in range(5):
+        repo.publish([Targets.type], verify_version=False)
+    repo.publish([Snapshot.type, Timestamp.type])
+
+    # Now the client should fail because it has version 7
+    # locally and the repository is presenting it with
+    # version 5
+    assert client.refresh(init_data) == 1
+
+    # Make sure the client still has version 7
+    assert client.version(Targets.type) == 7
 ```
 
 Running the test again results in a second repository version being dumped (each client refresh leads to a dump): 
@@ -133,28 +146,41 @@ The repository metadata looks as expected (but not spec-compliant) so we can add
 The final test looks like this:
 
 ```python
-def test_targets_version(
+def test_targets_version_rollback(
     client: ClientRunner, server: SimulatorServer
 ) -> None:
     # Initialize our repository: modify targets version and make sure it's included in snapshot
     init_data, repo = server.new_test(client.test_name)
-    repo.targets.version = 7
-    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
 
     # initialize client-under-test
     client.init_client(init_data)
-
     # Run refresh on client-under-test, expect success
     assert client.refresh(init_data) == 0
 
-    # Make a non-compliant change in repository
-    repo.targets.version = 6
-    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
+    # Bump version of Targest 6 times.
+    for i in range(6):
+        repo.publish([Targets.type])
+    repo.publish([Snapshot.type, Timestamp.type])
+    
+    # Client refreshes and downloads Targets version 7
+    assert client.refresh(init_data) == 0
+    assert client.version(Targets.type) == 7
 
-    # refresh client again, expect failure and refusal to accept snapshot and targets
+    # The repository deletes all published Targets
+    # and republishes to version 5.
+    # This is a rollback attack.
+    del repo.signed_mds[Targets.type]
+    for i in range(5):
+        repo.publish([Targets.type], verify_version=False)
+    repo.publish([Snapshot.type, Timestamp.type])
+
+    # Now the client should fail because it has version 7
+    # locally and the repository is presenting it with
+    # version 5
     assert client.refresh(init_data) == 1
-    assert client.version(Snapshot.type) == 3
-    assert client.version(Targets.type) == 8
+
+    # Make sure the client still has version 7
+    assert client.version(Targets.type) == 7
 ```
 
 ### Releasing
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -17,22 +17,19 @@ Example::
     # constructor creates repository with top-level metadata
     sim = RepositorySimulator()
 
-    # metadata can be modified directly: it is immediately available to clients
-    sim.snapshot.version += 1
+    # publish a new version of metadata
+    sim.root.expires = datetime.datetime.utcnow() + datetime.timedelta(days=7)
+    sim.publish([Root.type])
 
-    # As an exception, new root versions require explicit publishing
-    sim.publish("root", bump_version=True)
-
-    # there are helper functions
-    sim.add_target("targets", b"content", "targetpath")
-    sim.targets.version += 1
-    sim.update_snapshot()
+    # there are helper functions to do things like adding an artifact
+    sim.add_target(Targets.type, b"content", "targetpath")
+    sim.publish([Targets.type, Snapshot.type, Timestamp.type])
 """
 
 import datetime
 import logging
 import os
-from collections.abc import Iterable, Iterator
+from collections.abc import Iterable
 from dataclasses import dataclass
 from urllib import parse
 
@@ -48,7 +45,6 @@ from tuf.api.metadata import (
     Root,
     Snapshot,
     SuccinctRoles,
-    T,
     TargetFile,
     Targets,
     Timestamp,
@@ -154,13 +150,6 @@ class RepositorySimulator:
         assert isinstance(signed, Targets)
         return signed
 
-    def all_signed_targets(self) -> Iterator[tuple[str, Metadata[T]]]:
-        """Yield role name and signed portion of targets one by one."""
-        for role, mds in self.signed_mds.items():
-            for md in mds:
-                if role not in [Root.type, Timestamp.type, Snapshot.type]:
-                    yield role, Metadata.from_bytes(md).signed
-
     def new_signer(
         self, keytype: str = "rsa", scheme: str = "rsa-pkcs1v15-sha256"
     ) -> CryptoSigner:
@@ -228,6 +217,22 @@ class RepositorySimulator:
             if role not in self.signed_mds:
                 self.signed_mds[role] = []
             self.signed_mds[role].append(md.to_bytes(JSONSerializer()))
+
+            hashes = None
+            length = None
+            if self.compute_metafile_hashes_length:
+                hashes, length = self._compute_hashes_and_length(role)
+
+            # for targets roles: update snapshot.meta
+            if role not in [Root.type, Timestamp.type, Snapshot.type]:
+                self.snapshot.meta[f"{role}.json"] = MetaFile(
+                    md.signed.version, length, hashes
+                )
+            # for snapshot: update timestamp.snapshot_meta
+            if role == Snapshot.type:
+                self.timestamp.snapshot_meta = MetaFile(
+                    md.signed.version, length, hashes
+                )
 
     def fetch(self, path: str) -> bytes:
         """Fetches and returns metadata/artifacts for the given url-path.
@@ -310,34 +315,6 @@ class RepositorySimulator:
         digest_object.update(data)
         hashes = {sslib_hash.DEFAULT_HASH_ALGORITHM: digest_object.hexdigest()}
         return hashes, len(data)
-
-    def update_timestamp(self) -> None:
-        """Update timestamp and assign snapshot version to snapshot_meta
-        version.
-        """
-        hashes = None
-        length = None
-        if self.compute_metafile_hashes_length:
-            hashes, length = self._compute_hashes_and_length(Snapshot.type)
-
-        self.timestamp.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
-
-        self.publish([Timestamp.type])
-
-    def update_snapshot(self) -> None:
-        """Update snapshot, assign targets versions and update timestamp."""
-        for role, delegate in self.all_signed_targets():
-            hashes = None
-            length = None
-            if self.compute_metafile_hashes_length:
-                hashes, length = self._compute_hashes_and_length(role)
-
-            self.snapshot.meta[f"{role}.json"] = MetaFile(
-                delegate.version, length, hashes
-            )
-
-        self.publish([Snapshot.type])
-        self.update_timestamp()
 
     def add_artifact(self, role: str, data: bytes, path: str) -> None:
         """Add `data` to artifact store and insert its hashes into metadata."""
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -32,7 +32,7 @@ Example::
 import datetime
 import logging
 import os
-from collections.abc import Iterator
+from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
 from urllib import parse
 
@@ -59,8 +59,6 @@ from tuf_conformance.metadata import MetadataTest, RootTest
 logger = logging.getLogger(__name__)
 
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
-
-ALL_TOPLEVEL_TYPES = [Root.type, Timestamp.type, Snapshot.type, Targets.type]
 
 # Generate some signers once (to avoid all tests generating them)
 NUM_SIGNERS = 9
@@ -212,9 +210,7 @@ class RepositorySimulator:
             [Targets.type, Snapshot.type, Timestamp.type, Root.type], bump_version=False
         )
 
-    def publish(
-        self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = True
-    ) -> None:
+    def publish(self, roles: Iterable[str], bump_version: bool = True) -> None:
         for role in roles:
             md = self.mds.get(role)
             if md is None:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -342,7 +342,6 @@ class RepositorySimulator:
                 delegate.version, length, hashes
             )
 
-        self.snapshot.version += 1
         self.publish([Snapshot.type])
         self.update_timestamp()
 
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -194,6 +194,20 @@ class RepositorySimulator:
         self.publish([Targets.type, Snapshot.type, Timestamp.type, Root.type])
 
     def publish(self, roles: Iterable[str], bump_version: bool = True) -> None:
+        """Makes the repositorys metadata public and available to clients.
+
+        Tests run this helper after updating the repositorys metadata
+        and before clients should fetch the repositorys updated metadata.
+
+        publish bumps the version by default. The role must already exist
+        in repo.signed_mds for the repo to be able to bump it.
+
+        When publishing a role that is not root, timestamp or snapshot,
+        publish will create a metafile for the role and add it to snapshot.
+
+        When publishing snapshot, publish will create a metafile for snapshot
+        and add it to timestamp."""
+
         for role in roles:
             md = self.mds.get(role)
             if md is None:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -328,7 +328,6 @@ class RepositorySimulator:
 
         self.timestamp.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
 
-        self.timestamp.version += 1
         self.publish([Timestamp.type])
 
     def update_snapshot(self) -> None:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -206,8 +206,7 @@ class RepositorySimulator:
             signer = self.new_signer()
             self.root.add_key(signer.public_key, role)
             self.add_signer(role, signer)
-
-        self.publish([Root.type])
+        self.publish([Targets.type, Snapshot.type, Timestamp.type, Root.type])
 
     def publish(
         self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = False
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -300,8 +300,8 @@ class RepositorySimulator:
         # Non-root mds:
         if len(self.signed_mds[role]) == 0:
             raise ValueError(f"The repository has not published metadata for '{role}'")
-        # if version is not None:
-        #    return self.signed_mds[role][version - 1]
+        if version is not None:
+            return self.signed_mds[role][version - 1]
         return self.signed_mds[role][-1]
 
     def _version(self, role: str) -> int:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -21,8 +21,7 @@ Example::
     sim.snapshot.version += 1
 
     # As an exception, new root versions require explicit publishing
-    sim.root.version += 1
-    sim.publish_root()
+    sim.publish("root", bump_version=True)
 
     # there are helper functions
     sim.add_target("targets", b"content", "targetpath")
@@ -61,6 +60,8 @@ logger = logging.getLogger(__name__)
 
 SPEC_VER = ".".join(SPECIFICATION_VERSION)
 
+ALL_TOPLEVEL_TYPES = [Root.type, Timestamp.type, Snapshot.type, Targets.type]
+
 # Generate some signers once (to avoid all tests generating them)
 NUM_SIGNERS = 9
 SIGNERS = {
@@ -96,9 +97,8 @@ class RepositorySimulator:
         # All current metadata
         self.mds: dict[str, Metadata] = {}
 
-        # other metadata is signed on-demand (when fetched) but roots must be
-        # explicitly published with publish_root() which maintains this list
-        self.signed_roots: list[bytes] = []
+        # All signed metadata
+        self.signed_mds: dict[str, list[bytes]] = {}
 
         # signers are used on-demand at fetch time to sign metadata
         # keys are roles, values are dicts of {keyid: signer}
@@ -197,26 +197,44 @@ class RepositorySimulator:
         self.mds[Timestamp.type] = MetadataTest(Timestamp(expires=self.safe_expiry))
         self.mds[Root.type] = MetadataTest(RootTest(expires=self.safe_expiry))
 
+        self.signed_mds[Targets.type] = []
+        self.signed_mds[Snapshot.type] = []
+        self.signed_mds[Timestamp.type] = []
+        self.signed_mds[Root.type] = []
+
         for role in TOP_LEVEL_ROLE_NAMES:
             signer = self.new_signer()
             self.root.add_key(signer.public_key, role)
             self.add_signer(role, signer)
 
-        self.publish_root()
+        self.publish([Root.type])
 
-    def bump_root_by_one(self) -> None:
-        self.root.version += 1
-        self.publish_root()
+    def publish(
+        self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = False
+    ) -> None:
+        for r in roles:
+            role = parse.unquote(r, encoding="utf-8")
+            md = self.mds.get(role)
+            if md is None:
+                raise ValueError(f"Unknown role {role}")
 
-    def publish_root(self) -> None:
-        """Sign and store a new serialized version of root."""
-        root_md = self.mds[Root.type]
-        root_md.signatures.clear()
-        for signer in self.signers[Root.type].values():
-            root_md.sign(signer, append=True)
+            if bump_version:
+                md.signed.version += 1
 
-        self.signed_roots.append(root_md.to_bytes(JSONSerializer()))
-        logger.debug("Published root v%d", self.root.version)
+            md.signatures.clear()
+            for signer in self.signers[role].values():
+                md.sign(signer, append=True)
+
+            logger.debug(
+                "signed %s v%d with %d sigs",
+                role,
+                md.signed.version,
+                len(self.signers[role]),
+            )
+
+            if role not in self.signed_mds:
+                self.signed_mds[role] = []
+            self.signed_mds[role].append(md.to_bytes(JSONSerializer()))
 
     def fetch(self, path: str) -> bytes:
         """Fetches and returns metadata/artifacts for the given url-path.
@@ -274,33 +292,18 @@ class RepositorySimulator:
 
         If version is None, non-versioned metadata is being requested.
         """
+
         # decode role for the metadata
         role = parse.unquote(role, encoding="utf-8")
 
         if role == Root.type:
-            # return a version previously serialized in publish_root()
-            if version is None or version > len(self.signed_roots):
+            if version is None or version > len(self.signed_mds[Root.type]):
                 raise ValueError(f"Unknown root version {version}")
-            logger.debug("fetched root version %d", version)
-            return self.signed_roots[version - 1]
-
-        # sign and serialize the requested metadata
-        md = self.mds.get(role)
-
-        if md is None:
-            raise ValueError(f"Unknown role {role}")
-
-        md.signatures.clear()
-        for signer in self.signers[role].values():
-            md.sign(signer, append=True)
-
-        logger.debug(
-            "fetched %s v%d with %d sigs",
-            role,
-            md.signed.version,
-            len(self.signers[role]),
-        )
-        return md.to_bytes(JSONSerializer())
+            return self.signed_mds[Root.type][version - 1]
+        # Non-root mds:
+        if len(self.signed_mds[role]) == 0:
+            raise ValueError(f"The repository has not published metadata for '{role}'")
+        return self.signed_mds[role][-1]
 
     def _version(self, role: str) -> int:
         signed = self.mds[role].signed
@@ -308,7 +311,11 @@ class RepositorySimulator:
         return signed.version
 
     def _compute_hashes_and_length(self, role: str) -> tuple[dict[str, str], int]:
-        data = self.fetch_metadata(role)
+        md = self.mds[role]
+        md.signatures.clear()
+        for signer in self.signers[role].values():
+            md.sign(signer, append=True)
+        data = md.to_bytes(JSONSerializer())
         digest_object = sslib_hash.digest(sslib_hash.DEFAULT_HASH_ALGORITHM)
         digest_object.update(data)
         hashes = {sslib_hash.DEFAULT_HASH_ALGORITHM: digest_object.hexdigest()}
@@ -318,7 +325,6 @@ class RepositorySimulator:
         """Update timestamp and assign snapshot version to snapshot_meta
         version.
         """
-
         hashes = None
         length = None
         if self.compute_metafile_hashes_length:
@@ -425,16 +431,24 @@ class RepositorySimulator:
         dest_dir = os.path.join(self.dump_dir, f"refresh-{self.dump_version}")
         os.makedirs(dest_dir, exist_ok=True)
 
-        for ver in range(1, len(self.signed_roots) + 1):
+        for ver in range(1, len(self.signed_mds[Root.type]) + 1):
             with open(os.path.join(dest_dir, f"{ver}.root.json"), "wb") as f:
-                f.write(self.fetch_metadata(Root.type, ver))
+                try:
+                    data = self.fetch_metadata(Root.type, ver)
+                except ValueError:
+                    data = b"No metadata found"
+                f.write(data)
 
         for role in self.mds:
             if role == Root.type:
                 continue
             quoted_role = parse.quote(role, "")
             with open(os.path.join(dest_dir, f"{quoted_role}.json"), "wb") as f:
-                f.write(self.fetch_metadata(role))
+                try:
+                    data = self.fetch_metadata(role)
+                except ValueError:
+                    data = b"No metadata found"
+                f.write(data)
 
     def add_key(
         self,
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -48,6 +48,7 @@ from tuf.api.metadata import (
     Root,
     Snapshot,
     SuccinctRoles,
+    T,
     TargetFile,
     Targets,
     Timestamp,
@@ -153,7 +154,7 @@ class RepositorySimulator:
         assert isinstance(signed, Targets)
         return signed
 
-    def all_signed_targets(self) -> Iterator[tuple[str, Targets]]:
+    def all_signed_targets(self) -> Iterator[tuple[str, Metadata[T]]]:
         """Yield role name and signed portion of targets one by one."""
         for role, mds in self.signed_mds.items():
             for md in mds:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -157,9 +157,10 @@ class RepositorySimulator:
 
     def all_targets(self) -> Iterator[tuple[str, Targets]]:
         """Yield role name and signed portion of targets one by one."""
-        for role, md in self.mds.items():
-            if role not in [Root.type, Timestamp.type, Snapshot.type]:
-                yield role, md.signed
+        for role, mds in self.signed_mds.items():
+            for md in mds:
+                if role not in [Root.type, Timestamp.type, Snapshot.type]:
+                    yield role, Metadata.from_bytes(md).signed
 
     def new_signer(
         self, keytype: str = "rsa", scheme: str = "rsa-pkcs1v15-sha256"
@@ -337,7 +338,6 @@ class RepositorySimulator:
     def update_snapshot(self) -> None:
         """Update snapshot, assign targets versions and update timestamp."""
         for role, delegate in self.all_targets():
-            self.publish([role])
             hashes = None
             length = None
             if self.compute_metafile_hashes_length:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -437,22 +437,14 @@ class RepositorySimulator:
 
         for ver in range(1, len(self.signed_mds[Root.type]) + 1):
             with open(os.path.join(dest_dir, f"{ver}.root.json"), "wb") as f:
-                try:
-                    data = self.fetch_metadata(Root.type, ver)
-                except ValueError:
-                    data = b"No metadata found"
-                f.write(data)
+                f.write(self.fetch_metadata(Root.type, ver))
 
-        for role in self.mds:
+        for role in self.signed_mds:
             if role == Root.type:
                 continue
             quoted_role = parse.quote(role, "")
             with open(os.path.join(dest_dir, f"{quoted_role}.json"), "wb") as f:
-                try:
-                    data = self.fetch_metadata(role)
-                except ValueError:
-                    data = b"No metadata found"
-                f.write(data)
+                f.write(self.fetch_metadata(role))
 
     def add_key(
         self,
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -155,7 +155,7 @@ class RepositorySimulator:
         assert isinstance(signed, Targets)
         return signed
 
-    def all_targets(self) -> Iterator[tuple[str, Targets]]:
+    def all_signed_targets(self) -> Iterator[tuple[str, Targets]]:
         """Yield role name and signed portion of targets one by one."""
         for role, mds in self.signed_mds.items():
             for md in mds:
@@ -334,7 +334,7 @@ class RepositorySimulator:
 
     def update_snapshot(self) -> None:
         """Update snapshot, assign targets versions and update timestamp."""
-        for role, delegate in self.all_targets():
+        for role, delegate in self.all_signed_targets():
             hashes = None
             length = None
             if self.compute_metafile_hashes_length:
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -329,8 +329,7 @@ class RepositorySimulator:
         if self.compute_metafile_hashes_length:
             hashes, length = self._compute_hashes_and_length(Snapshot.type)
 
-        md = self.mds[Timestamp.type]
-        md.signed.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
+        self.timestamp.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
 
         self.timestamp.version += 1
         self.publish([Timestamp.type])
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -301,7 +301,19 @@ class RepositorySimulator:
     def fetch_metadata(self, role: str, version: int | None = None) -> bytes:
         """Return signed metadata for 'role', using 'version' if it is given.
 
-        If version is None, non-versioned metadata is being requested.
+        If version is None, non-versioned metadata is being requested, and the
+        repository will return the metadata it has published last.
+
+        fetch_metadata versions its metadata by storing the version in
+        repo.signed_mds[role][version-1].
+
+        For the metadata to be available to the client, it must exist in
+        repo.signed_mds. repo.publish() is responsible for makeing the
+        metadata available in repo.signed_mds. The common workflow for
+        making changes in a test and fetching is:
+        1. Make changes to the metadata in the test
+        2. Publish the metadata by way of repo.publish([role])
+        3. The client can now fetch the updated metadata.
         """
 
         # decode role for the metadata
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -196,19 +196,12 @@ class RepositorySimulator:
         self.mds[Timestamp.type] = MetadataTest(Timestamp(expires=self.safe_expiry))
         self.mds[Root.type] = MetadataTest(RootTest(expires=self.safe_expiry))
 
-        self.signed_mds[Targets.type] = []
-        self.signed_mds[Snapshot.type] = []
-        self.signed_mds[Timestamp.type] = []
-        self.signed_mds[Root.type] = []
-
         for role in TOP_LEVEL_ROLE_NAMES:
             signer = self.new_signer()
             self.root.add_key(signer.public_key, role)
             self.add_signer(role, signer)
 
-        self.publish(
-            [Targets.type, Snapshot.type, Timestamp.type, Root.type], bump_version=False
-        )
+        self.publish([Targets.type, Snapshot.type, Timestamp.type, Root.type])
 
     def publish(self, roles: Iterable[str], bump_version: bool = True) -> None:
         for role in roles:
@@ -216,7 +209,8 @@ class RepositorySimulator:
             if md is None:
                 raise ValueError(f"Unknown role {role}")
 
-            if bump_version:
+            # only bump if there is a version already (this avoids bumping initial v1)
+            if bump_version and role in self.signed_mds:
                 md.signed.version += 1
 
             md.signatures.clear()
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -207,7 +207,10 @@ class RepositorySimulator:
             signer = self.new_signer()
             self.root.add_key(signer.public_key, role)
             self.add_signer(role, signer)
-        self.publish([Targets.type, Snapshot.type, Timestamp.type, Root.type])
+
+        self.publish(
+            [Targets.type, Snapshot.type, Timestamp.type, Root.type], bump_version=False
+        )
 
     def publish(
         self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = True
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -210,7 +210,7 @@ class RepositorySimulator:
         self.publish([Targets.type, Snapshot.type, Timestamp.type, Root.type])
 
     def publish(
-        self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = False
+        self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = True
     ) -> None:
         for role in roles:
             md = self.mds.get(role)
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -312,9 +312,6 @@ class RepositorySimulator:
 
     def _compute_hashes_and_length(self, role: str) -> tuple[dict[str, str], int]:
         md = Metadata.from_bytes(self.signed_mds[role][-1])
-        md.signatures.clear()
-        for signer in self.signers[role].values():
-            md.sign(signer, append=True)
         data = md.to_bytes(JSONSerializer())
         digest_object = sslib_hash.digest(sslib_hash.DEFAULT_HASH_ALGORITHM)
         digest_object.update(data)
--- a/tuf_conformance/repository_simulator.py
+++ b/tuf_conformance/repository_simulator.py
@@ -311,11 +311,8 @@ class RepositorySimulator:
         If version is None, non-versioned metadata is being requested, and the
         repository will return the metadata it has published last.
 
-        fetch_metadata versions its metadata by storing the version in
-        repo.signed_mds[role][version-1].
-
         For the metadata to be available to the client, it must exist in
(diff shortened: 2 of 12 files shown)


jku (changes requested):
It does look as annoying change as expected -- reviewing the test cases is going to be tricky so I'm going to try and cover all of the "architecture" decisions first.

It's entirely possible I just don't see why some of the suggestions are unworkable, just push back if so.

### `update_timestamp()` and `updated_snapshot`

* these methods are now incorrect: the data that goes into `timestamp.snapshot_meta` and `snapshot.meta` **must be based on signed/published metadata**. The tricks you end up doing in _compute_hashes_and_length() are a symptom of this issue and would fail if you tried them with e.g. ecdsa sigs since the hashes are now based on completely different data than the one we send to clients
* the logical followup on that is that **tests must publish in correct order**: e.g. always publish targets before `update_snapshot()`. Same applies to any manual changes to snapshot: they must be published before `update_timestamp()`. Basically all calls like this `repo.publish([Timestamp.type, Snapshot.type, Targets.type])` are very suspect: you cannot first publish timestamp, then snapshot, then targets.
* these methods should IMO `publish()` internally (and I think that neatly solves the above problem). This will remove loads of added `publish()` calls from tests
* if you think the naming is not clear enough (in that it's not obvious that they also publish), I'm fine with renaming ... but they seem fine to me

### bump_version default

publish() should bump version by default. The only major case where you did not set `bump_version=True` was after update_timestamp/update_snapshot -- and as mentioned in previous section I believe those must be changed to handle publishing internally anyway. 

### simulator initialization

We should just initialize all top level roles (and not just root) when initializing: 
* this feels intuitive: a valid repository is usually the right starting point
* then you don't need to add the `publish_all` argument to new_test() if you just fix the two tests that expect something else

jku:
> * he logical followup on that is that **tests must publish in correct order**: e.g. always publish targets before `update_snapshot()`. Same applies to any manual changes to snapshot: they must be published before `update_timestamp()`. Basically all calls like this `repo.publish([Timestamp.type, Snapshot.type, Targets.type])` are very suspect: you cannot first publish timestamp, then snapshot, then targets.

And final followup: because of the above, there is probably no point in the publish() argument being a list. I think there will be very few cases where that is useful -- but I guess well find out as the PR progresses

jku:
We should switch the default key back to ecdsa in this same PR: that will validate that the hash calculation is done on the same data we send to clients

tuf_conformance/repository_simulator.py
```diff
@@ -274,41 +292,30 @@ def fetch_metadata(self, role: str, version: int | None = None) -> bytes:
...
-            role,
-            md.signed.version,
-            len(self.signers[role]),
-        )
-        return md.to_bytes(JSONSerializer())
+            return self.signed_mds[Root.type][version - 1]
+        # Non-root mds:
+        if len(self.signed_mds[role]) == 0:
+            raise ValueError(f"The repository has not published metadata for '{role}'")
+        return self.signed_mds[role][-1]
```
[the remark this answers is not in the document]
AdamKorcz: The repository gets a lot of calls for non-versioned, non-root roles. The only way I have made it work is to return the latest version for non-root roles. I haven't found a way to make it work when we return `self.signed_mds[role][version - 1]`. On the same note, since `Root` throws an error if the caller does not specify a version, and non-root roles don't, I have also not found a way to merge the two paths.

AdamKorcz:
> update_timestamp/update_snapshot() should IMO call publish() internally (and I think that neatly solves the above problem). This will remove loads of added publish() calls from tests

I wonder what is smartest here; 1) that `update_timestamp` and `update_snapshot` publish, or 2) that `publish` adopts the logic that is currently in `update_timestamp` and `update_snapshot`. It might be a UX-upgrade if we only have `publish` instead of `update_timestamp`/`update_snapshot`/`publish` that all more or less do the same?

AdamKorcz:
> And final followup: because of the above, there is probably no reason for the publish() argument to be a list. I think there will be very few cases where that is useful -- but I guess well find out as the PR progresses

We should explore if we can always publish timestamp, snapshot and targets. I.e. everytime a tests wants to publish one of these or a custom role, does it also want to publish timestamp, snapshot and targets? If so, we can reduce the need for specifying these which should make the UX simpler too.

AdamKorcz:
> publish() should bump version by default. The only major case where you did not set bump_version=True was after update_timestamp/update_snapshot -- and as mentioned in previous section I believe those must be changed to handle publishing internally anyway.

You mean we can bump the version of timestamp/snapshot/targets per default? There might be some cases that don't need that:

https://github.com/theupdateframework/tuf-conformance/pull/188/files#diff-876f048f89628d9ed185a0965b6fc8a1544f5e85515a75e2ac8b3d2d4fd67f88R60-R61

https://github.com/theupdateframework/tuf-conformance/pull/188/files#diff-876f048f89628d9ed185a0965b6fc8a1544f5e85515a75e2ac8b3d2d4fd67f88R215-R216

https://github.com/theupdateframework/tuf-conformance/pull/188/files#diff-876f048f89628d9ed185a0965b6fc8a1544f5e85515a75e2ac8b3d2d4fd67f88R130-R132

tuf_conformance/repository_simulator.py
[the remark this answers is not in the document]
jku: There is two issues here: 
* the fixing the separate code paths is a suggestion, feel free to not do it
* We still should not answer with `self.signed_mds[role][-1]` _when the caller is asking for a specific version_. This is a bug


> since Root throws an error if the caller does not specify a version, and non-root roles don't

I don't think we need to raise in that root case, we can probably just make it work like the other roles: return the latest version

jku:
> I wonder what is smartest here; 1) that `update_timestamp` and `update_snapshot` publish, or 2) that `publish` adopts the logic that is currently in `update_timestamp` and `update_snapshot`. It might be a UX-upgrade if we only have `publish` instead of `update_timestamp`/`update_snapshot`/`publish` that all more or less do the same?

A real repository would indeed have a `publish()` that just does the right thing (updates snapshot, updates timestamp, etc). But in our case we want the test to have control over what happens. We want it to be almost as easy to do the wrong thing than the right thing. I don't think we can hide the details (like `update_snapshot()`) from the tests without making it tricky to build broken repositories.


I think our `publish()` should not be smart, it should just make the given metadata available to client -- the only exception I believe makes sense is that it should bump version because that is something we want 95% of the time

Can we add arguments to `publish()` so it optionally updates snapshot too? I guess so but I think it will make this PR more complicated, and it's not clear to me that it's easier to understand.

tuf_conformance/repository_simulator.py
```diff
@@ -197,26 +197,44 @@ def _initialize(self) -> None:
...
-        """Sign and store a new serialized version of root."""
-        root_md = self.mds[Root.type]
-        root_md.signatures.clear()
-        for signer in self.signers[Root.type].values():
-            root_md.sign(signer, append=True)
+    def publish(
+        self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = False
+    ) -> None:
+        for r in roles:
+            role = parse.unquote(r, encoding="utf-8")
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/fc8f74c6780fbf2e918a7d1a02c334e026ca943e

tuf_conformance/simulator_server.py
```diff
@@ -61,7 +63,9 @@ def __init__(self, dump_dir: str | None) -> None:
         # key is test name, value is the repository sim for that test
         self.repos: dict[str, RepositorySimulator] = {}
 
-    def new_test(self, name: str) -> tuple[ClientInitData, RepositorySimulator]:
+    def new_test(
+        self, name: str, publish_all: bool = True
+    ) -> tuple[ClientInitData, RepositorySimulator]:
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/ef4f2dfdda452a811d0663dffb28e67040ce711d

tuf_conformance/repository_simulator.py
```diff
@@ -96,9 +97,8 @@ def __init__(self, dump_dir: str | None) -> None:
         # All current metadata
         self.mds: dict[str, Metadata] = {}
 
-        # other metadata is signed on-demand (when fetched) but roots must be
-        # explicitly published with publish_root() which maintains this list
-        self.signed_roots: list[bytes] = []
+        # All signed metadata
+        self.signed_mds: dict[str, list[bytes]] = {}
 
         # signers are used on-demand at fetch time to sign metadata
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/b3c46865519e3e07aad0a6af0a90e2de18ced56e

tuf_conformance/repository_simulator.py
```diff
@@ -274,41 +292,30 @@ def fetch_metadata(self, role: str, version: int | None = None) -> bytes:
...
         assert isinstance(signed, Root | Timestamp | Snapshot | Targets)
         return signed.version
 
     def _compute_hashes_and_length(self, role: str) -> tuple[dict[str, str], int]:
-        data = self.fetch_metadata(role)
+        md = self.mds[role]
+        md.signatures.clear()
+        for signer in self.signers[role].values():
+            md.sign(signer, append=True)
+        data = md.to_bytes(JSONSerializer())
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/d3044ffb83fa4ee0f44d30a27e36931ae08fd2c8

tuf_conformance/repository_simulator.py
```diff
@@ -425,16 +435,24 @@ def debug_dump(self) -> None:
...
         for role in self.mds:
             if role == Root.type:
                 continue
             quoted_role = parse.quote(role, "")
             with open(os.path.join(dest_dir, f"{quoted_role}.json"), "wb") as f:
-                f.write(self.fetch_metadata(role))
+                try:
+                    data = self.fetch_metadata(role)
+                except ValueError:
+                    data = b"No metadata found"
```
jku, line 454: come on, there is clearly a bug somewhere if this happens. Let's not hide that.

Looping through `self.signed_mds` instead of `self.mds` should be fine

tuf_conformance/simulator_server.py
```diff
@@ -79,6 +81,10 @@ def new_test(self, name: str) -> tuple[ClientInitData, RepositorySimulator]:
             repo.fetch_metadata("root", 1),
         )
 
+        # The repository simulator publishes the root.
+        # Here, we publish the remaining metadata:
+        repo.publish([Targets.type, Snapshot.type, Timestamp.type])
```
jku, line 86: I would like this to happen in the same place where root is published or to have a a reason why it is hidden in the http server.

tuf_conformance/repository_simulator.py
```diff
@@ -197,26 +197,43 @@ def _initialize(self) -> None:
...
 
-    def publish_root(self) -> None:
-        """Sign and store a new serialized version of root."""
-        root_md = self.mds[Root.type]
-        root_md.signatures.clear()
-        for signer in self.signers[Root.type].values():
-            root_md.sign(signer, append=True)
+    def publish(
+        self, roles: list[str] = ALL_TOPLEVEL_TYPES, bump_version: bool = False
+    ) -> None:
```
jku, line 214: I still believe bump_version should be True by default

tuf_conformance/test_repository_simulator.py
```diff
@@ -0,0 +1,81 @@
...
+            expected_length,
+        )
+        self.assertEqual(
+            Metadata.from_bytes(
+                repo.signed_mds[Timestamp.type][-1]
+            ).signed.snapshot_meta.hashes,
+            expected_hashes,
+        )
+
+    def test_timestamp_expired_unittest(self) -> None:
```
jku, line 66: maybe rename to `test_update_timestamp`

jku (changes requested):
I have not started reviewing the actual tests yet, would like the infra too look right first as otherwise the test results are not right yet.

I expect there to be quite a bit more changes required in tests: every single time metadata is modified, it now needs to be published. Can you document whether you have reviewed every test looking for this or just handled test failures?

jku:
Oh, on test_repository_simulator.py, I do like the idea of unit tests but:
* It's a bit weird to use unittest if the conformance test suite uses pytest
* I think these tests should live in a separate directory and the actual conformance test suite should not run tests like this, only our own CI should

tuf_conformance/test_repository_simulator.py
```diff
@@ -0,0 +1,81 @@
...
+        self.assertEqual(
+            Metadata.from_bytes(repo.signed_mds[Snapshot.type][-1]).signed.version, 2
+        )
+
+        expected_hashes, expected_length = repo._compute_hashes_and_length(
+            Snapshot.type
+        )
+
+        self.assertEqual(
+            Metadata.from_bytes(
```
jku, line 54: You can add the missing annotation and silence mypy with

`Metadata[Timestamp].from_bytes(`

AdamKorcz:
>I expect there to be quite a bit more changes required in tests: every single time metadata is modified, it now needs to be published (with either publish() or update_timestamp/update_snapshot()). Can you document whether you have reviewed every test looking for this or just handled test failures?

I have not. Note that this will contradict with the version lookup, since we append to `signed_mds` every time we publish. So if we publish _every time_ we make a change to the metadata, we can append multiple versions with the same version number.

tuf_conformance/repository_simulator.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/ebb56b2be921fa03803750c7487a3c5d5f512f31

tuf_conformance/simulator_server.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/2caf72a241205ef262c5ee2fd58efcfb665b98c6

tuf_conformance/test_repository_simulator.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/984b4abfb0ee3f797268031d19c97be29922493c

tuf_conformance/repository_simulator.py
```diff
@@ -329,13 +330,16 @@ def update_timestamp(self) -> None:
         if self.compute_metafile_hashes_length:
             hashes, length = self._compute_hashes_and_length(Snapshot.type)
 
-        self.timestamp.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
+        md = self.mds[Timestamp.type]
+        md.signed.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/075463bdb6ef20b6a21a2ad4e506e17c1f5f63c5

tuf_conformance/repository_simulator.py
```diff
@@ -274,41 +290,32 @@ def fetch_metadata(self, role: str, version: int | None = None) -> bytes:
...
-            md.signed.version,
-            len(self.signers[role]),
-        )
-        return md.to_bytes(JSONSerializer())
+            return self.signed_mds[Root.type][version - 1]
+        # Non-root mds:
+        if len(self.signed_mds[role]) == 0:
+            raise ValueError(f"The repository has not published metadata for '{role}'")
+        # if version is not None:
+        #    return self.signed_mds[role][version - 1]
```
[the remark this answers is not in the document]
AdamKorcz: I mention this here as well: https://github.com/theupdateframework/tuf-conformance/pull/188#issuecomment-2349364998, but essentially, this is only true if we have one version of each version number in the published metadata (`repo.signed_mds`), and currently it looks like the tests will not follow that paradigm. Do we need the published versions to be a dict instead that contains the versions as keys, and when we publish to an existing version, we overwrite that version? For example:

`self.signed_mds: dict[str (role), dict[str (version), list[bytes]]] = {}`

AdamKorcz:
> we could merge test_repository_simulator.py as is if it gets moved before next release... or drop it from this PR before merging and make that a separate PR

Either way works for me. I can also move it to another dir in this PR.

jku:
> So if we publish every time we make a change to the metadata, we can append multiple versions of a given role with the same version number.

I meant `"we need to publish everytime that we want changes to be visible to clients"` -- there are definitely tests where metadata is modified but is then never published. This is fine if the tests have not been reviewed yet.

tuf_conformance/test_file_download.py
```diff
@@ -234,7 +259,11 @@ def test_download_with_unknown_hash_algorithm(
     del target.hashes["sha512"]
     repo.targets.targets[target_path] = target
     repo.artifacts[target_path] = Artifact(target_content, target)
+    repo.timestamp.version += 1
+    repo.snapshot.version += 1
+    repo.targets.version += 1
     repo.publish([Targets.type, Snapshot.type, Timestamp.type])
+    repo.update_snapshot()
```
jku, line 266: I won't reiew tests yet but as an example of what I expect to happen once the version bumping happens by default:     

```
repo.publish(Targets.type)
repo.update_snapshot()
```

that should just work

jku (commented):
Can you update the status (of the infrastructure side) here?

I have suggested bump_version should default to True in `publish()`several times but I've not seen you respond to it. But I see tests being updated so they manually bump all versions... Do you disagree with the idea of bumping version by default or is this WIP?

tuf_conformance/repository_simulator.py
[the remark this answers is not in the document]
AdamKorcz: Fixed in https://github.com/theupdateframework/tuf-conformance/pull/188/commits/edea8b4580f204396eb8a787274648bfe4e6e19e

tuf_conformance/repository_simulator.py
```diff
@@ -329,13 +330,16 @@ def update_timestamp(self) -> None:
...
+        md = self.mds[Timestamp.type]
+        md.signed.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
 
         self.timestamp.version += 1
+        self.publish([Timestamp.type])
 
     def update_snapshot(self) -> None:
         """Update snapshot, assign targets versions and update timestamp."""
         for role, delegate in self.all_targets():
+            self.publish([role])
```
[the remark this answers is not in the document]
AdamKorcz: Removed in https://github.com/theupdateframework/tuf-conformance/pull/188/commits/49a08739b2afbc4f8b5d9e0adf846700b078c4ba#diff-14194f3ec16d9cbd8ebf0cc25b2a6a6040e4a56e620203f3373fd6cf426f946a

tuf_conformance/repository_simulator.py
```diff
@@ -310,7 +311,7 @@ def _version(self, role: str) -> int:
         return signed.version
 
     def _compute_hashes_and_length(self, role: str) -> tuple[dict[str, str], int]:
-        md = self.mds[role]
+        md = Metadata.from_bytes(self.signed_mds[role][-1])
         md.signatures.clear()
         for signer in self.signers[role].values():
             md.sign(signer, append=True)
```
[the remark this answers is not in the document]
AdamKorcz: Removed in https://github.com/theupdateframework/tuf-conformance/pull/188/commits/67a8290c3136a5364a70668d743b7e8510e04ef1

AdamKorcz:
> Can you update the status (of the infrastructure side) here?

Sure. All in all, I believe all changes have been addressed except for https://github.com/theupdateframework/tuf-conformance/pull/188#discussion_r1743768802, but let's put a pin in that. I think there were a few bugs that prevented this from working as intended for now.
Currently the PR reflects the following infrastructure changes:

1. The repo now has a `signed_mds` that has all the signed and published metadata. This is the metadata that clients receive when they fetch. It includes both top-level roles and other roles as well. We have discussed matching published versions by its index in the list of published metadata, and I think we should proceed with that. We can make it work. As long as we keep each item in the list a seperate version, we should be fine. The biggest hurdles I can think of here are:
1.a. The rollback tests that set a version of 9999 and then release version 1. I suspect we may need 9999 items in the list to do this.
1.b. We may need to bump some metadata versions more than we currently do in some tests. This should be fine, as the tests will be able to test for the same.
2. The PR adds the `repo.publish()` method which takes an MD from `repo.mds`, signs it and puts it in `repo.signed_mds`.
2.a. Remove `bump_root_by_one` and `publish_root` and essentially handle the same logic in `repo.publish()`
3. The repository now publishes all top-level roles when it is initialized. I believe a single test has to work around that, and it is easy.
4. `update_snapshot` and `update_timestamp` are important helper methods for many tests. I think most tests will use these instead of or in addition to manually invoking `repo.publish()`.

On the question of whether `repo.publish()` should bump the version by default: I have been undecided and have been going through the tests to see the impact in terms of UX. I have made the current required changes very clear in some tests to display the necessary changes. Prior to that I was undecided. AFAICT, we are currently dealing with the following problems in tests here:
1. There are tests that will need to bump the version for the clients to see the changes. When they `publish`, they need to bump the MD version and also run `repo.update_snapshot()`. Otherwise the test breaks. We will also need to change the assertions for MD versions in these tests. These tests should still be able to test for the same as they do now with the changes we need to make. 
2. There are tests that don't need to bump the version of non-root metadata to work and test for the same they test for. I believe these tests would not suffer from bumping the versions of all top-level roles; We would still be able to understand what they test for, but there would be some underlying assumptions about the metadata versions, for example the test publishes 3 times and then we expect the MD version to be 4.
3. There are tests that need to not bump the version. This should be fine if bumping is true by default but still optional.
4. From what I can tell, the balance is roughly 45% 1.a., 45% 1.b. and 10% 1.c. I think the biggest problem in bumping by default is 1) for everyone who views the tests to still understand the tests and 2) that it is clear when a test needs to bump, when it doesn't and when it should use `repo.update_snapshot`. This can be solved with documentation IMO.

Based on the above, I think we can bump by default. I have done that in [XXX](https://github.com/theupdateframework/tuf-conformance/pull/188/commits/12bb74bf50dd3ecfca72a348a36c0080ed423a1a). The tests that bump manually need a run-through.

tuf_conformance/repository_simulator.py
```diff
@@ -157,9 +157,10 @@ def any_targets(self, role: str) -> Targets:
 
     def all_targets(self) -> Iterator[tuple[str, Targets]]:
         """Yield role name and signed portion of targets one by one."""
-        for role, md in self.mds.items():
-            if role not in [Root.type, Timestamp.type, Snapshot.type]:
-                yield role, md.signed
+        for role, mds in self.signed_mds.items():
+            for md in mds:
+                if role not in [Root.type, Timestamp.type, Snapshot.type]:
+                    yield role, Metadata.from_bytes(md).signed
```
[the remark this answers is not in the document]
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/97701cb0d896da2218de068c48f67829cf412062

tuf_conformance/test_file_download.py
AdamKorcz: I agree. The purpose of that change was to make it clear why we need to bump by default.

tuf_conformance/repository_simulator.py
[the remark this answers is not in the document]
AdamKorcz: I will get back to this. Keeping it open for now.

jku:
Thanks! On the open questions:

> 1.a. The rollback tests that set a version of 9999 and then release version 1. I suspect we may need 9999 items in the list to do this.

Sure, sort of:
* we can of course lower the value of 9999 to e.g. 2 or 3...
* but we also don't _really_ need to generate all versions between one and 9999: the test will never ask for them so as long as `signed_mds[role]` contains an array with length 9999 and some bytes in the correct slot  we'd be fine

I guess option one is simpler even if it does not make the example of "fast-forward recovery" as clear...

These fast forward recovery tests are also be special in the sense that we have to "wipe" `signed_mds[role]` when the recovery happens (or at least write to `signed_mds[role][0]`).

> 1.b. We may need to bump some metadata versions more than we currently do in some tests. This should be fine, as the tests will be able to test for the same.

yes, agreed

AdamKorcz:
Great. Would you mind giving the repository simulator another review?

tuf_conformance/repository_simulator.py
```diff
@@ -21,8 +21,7 @@
     sim.snapshot.version += 1
 
     # As an exception, new root versions require explicit publishing
-    sim.root.version += 1
-    sim.publish_root()
+    sim.publish("root", bump_version=True)
```
jku, line 24: The example needs a bit of love :
* bump_version is not needed in this publish call 
* the example shows how root and snapshot changes are done differently: that's not true or useful anymore -- could remove one of those changes maybe?
*  `publish([Targets.type])` should be used instead of `sim.targets.version += 1`

tuf_conformance/repository_simulator.py
```diff
@@ -155,11 +155,12 @@ def any_targets(self, role: str) -> Targets:
...
-    def all_targets(self) -> Iterator[tuple[str, Targets]]:
+    def all_signed_targets(self) -> Iterator[tuple[str, Targets]]:
         """Yield role name and signed portion of targets one by one."""
-        for role, md in self.mds.items():
-            if role not in [Root.type, Timestamp.type, Snapshot.type]:
-                yield role, md.signed
+        for role, mds in self.signed_mds.items():
+            for md in mds:
+                if role not in [Root.type, Timestamp.type, Snapshot.type]:
+                    yield role, Metadata.from_bytes(md).signed
```
jku, line 163: hmm, rereading this, it now yields _every version_ of every targets metadata. I think what we want is _current version_ of every targets metadata

This probably already has correct test results since the only call site will end up rewriting the same variables for each version so always uses the current version last...  so we could leave it as is and file an issue to refactor the whole method away later (by storing hashes/length when targets roles are published).

But removing the inner loop is fine too

tuf_conformance/repository_simulator.py
```diff
@@ -274,41 +290,29 @@ def fetch_metadata(self, role: str, version: int | None = None) -> bytes:
...
 
     def _version(self, role: str) -> int:
         signed = self.mds[role].signed
         assert isinstance(signed, Root | Timestamp | Snapshot | Targets)
         return signed.version
 
     def _compute_hashes_and_length(self, role: str) -> tuple[dict[str, str], int]:
-        data = self.fetch_metadata(role)
+        md = Metadata.from_bytes(self.signed_mds[role][-1])
+        data = md.to_bytes(JSONSerializer())
```
jku, line 315: deserialization is not needed  and is in fact actively harmful (since we want to be able to put invalid data into `signed_mds[role]`).

```suggestion
        data = self.signed_mds[role][-1]
```

tuf_conformance/repository_simulator.py
```diff
@@ -327,10 +330,11 @@ def update_timestamp(self) -> None:
         self.timestamp.snapshot_meta = MetaFile(self.snapshot.version, length, hashes)
 
         self.timestamp.version += 1
```
jku, line 332: unneeded version bump

tuf_conformance/repository_simulator.py
```diff
@@ -341,6 +345,7 @@ def update_snapshot(self) -> None:
             )
 
         self.snapshot.version += 1
```
jku, line 347: unneeded version bump

jku (changes requested):
There's a couple of places that look like bugs but after those I think we are ready to handle tests...

I can take some test files to update once the simulator is ready

tuf_conformance/repository_simulator.py
AdamKorcz: Let's revisit this once we get to the tests after the repo is set up nicely. Keeping it open.

tuf_conformance/repository_simulator.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/caae1c1395d88221e02c445074289497256a0287

tuf_conformance/repository_simulator.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/383e6dae05b2701eab6a2a8e59de1a1dfa0886fa

tuf_conformance/repository_simulator.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/fd4d96b7394a4fff6e68d9beed2e7d81ae973bf1

AdamKorcz:
> I can take some test files to update once the simulator is ready

Help on the `test_updater_delegation_graphs.py` would be much appreciated.

jku:
I'll push a commit to fix delegation tests.

Note that I'm fixing a couple of things in RepositorySimulator as well -- this will change things in other tests  too but I assume for the better. Feel free to remove my commits and fix otherwise if this is a problem for your work.

jku:
The changes "fix" 47 tests so current tally is `31 failed, 63 passed`.

However, I have only reviewed tests in `test_updater_delegation_graphs.py` so far

jku:
Fixed quoting test as well and one more tweak to the simulator infra: I added a special case for initial version (so initial version of every role actually is 1).

Again, I can make these change as PRs to this branch or something else as well if you'd prefer that.

I'm not planning to work on other tests today. Current situation is:

**25 failed, 69 passed**
Reviewed:
* [x] test_quoting_issues.py
* [x] test_updater_delegation_graphs.py

jku:
> so we could ... file an issue to refactor the whole method away later (by storing hashes/length when targets roles are published).

I think this is likely the best way to make the metadata hash support tests simple so might still make sense in this PR:
* when targets metadata is published, store the hashes/length/version
* when snapshot needs to be updated, use the stored values
* tests that need to do shenanigans can just modify the data structure after publishing but before updating snapshot

I *think* we don't need to add any new data structures, publishing targets metadata can likely just modify `repo.snapshot.meta` directly -- I think that might be ok because it still makes it possible for tests to manually modify `repo.snapshot.meta` after publishing targets metadata but before snapshot is published.  If that is possible then as next step we can see if `update_snapshot()` really could be replaced with `publish([Snapshot.type, Timestamp.type])` like you suggested at some point. This might be out of scope for this PR though.

I'll have a look at the tests using `compute_metafile_hashes_length` in the next days and propose something

AdamKorcz:
Current status is:

**9 failed, 85 passed**

`test_non_root_rotations` is responsible for 7 of those failed.

jku:
Can you please document if you reviewed the tests/files or just got them green?

AdamKorcz:
> Can you please document if you reviewed the tests/files or just got them green?

Both. I have made them green in a way that they IMO still test the intended problem, but there may be a way to further reduce the tests. That being said, I think they need another round of review and be scrutinized for whether:

1. still test for the same.
2. can be reduced further.

jku:
I did the change we discussed, it turned out to be simple: 
* update_snapshot() and update_timestamp() are no more
* Instead new snapshot and timestamp are published like all other roles
* The "meta" changes are now done like this:
    * when any targets role is published, self.snapshot.meta is updated
    * when snapshot is published, self.timestamp.snapshot_meta is updated
    
This leads to the normal use case (where meta should match the published
metadata):

    repo.publish([Targets.type, Snapshot.type, Timestamp.type])

and the sneaky use case (where the meta is modified so it does not match):

    repo.publish([Targets.type])
    # modify repo.snapshot.meta here
    repo.publish([Snapshot.type, Timestmap.type])

This seems to be fairly logical and allows us to mostly not modify the test repositories -- the tests are almost all passing even though I did almost nothing to actually fix them: I just fixed the code to look right.

So this looks good to me: 
* My latest changes are ready for review
* I think we can now start reviewing the tests and fixing whatever still looks wrong
* we should try to look at the repository dump changes to verify that we are not changing tests that should not be changed -- but that could happen after everything else is done

jku:
The failing  `test_new_snapshot_version_mismatch()` does not test what it claims to test:
* it says  _Tests that the client does not download the snapshot metadata if the repo has bumped the snapshot version in the snapshot metadata but not in timestamp.meta._ This does not make a whole lot of sense since if timestamp is not updated ... then how would the client even know to download the new snapshot? We can test it though, no problem
* however, what the test actually does in main is this:
  * when client requests snapshot v1, repository returns snapshot v2
  * This is a good idea but seems accidental since the docstring describes this so differently

tuf_conformance/test_basic.py
```diff
@@ -173,22 +178,25 @@ def test_basic_metadata_hash_support(
 
     # Construct repository with hashes in timestamp/snapshot
     repo.compute_metafile_hashes_length = True
-    repo.update_snapshot()  # v2
+    repo.publish([Targets.type])  # v2
+    repo.publish([Snapshot.type, Timestamp.type])
```
AdamKorcz, line 182: We should combine these two `repo.publish` calls.

tuf_conformance/test_expiration.py
```diff
@@ -116,14 +116,14 @@ def test_expired_local_root(client: ClientRunner, server: SimulatorServer) -> No
     # root v2 expires in 7 days
     now = datetime.datetime.now(timezone.utc)
     repo.root.expires = now + datetime.timedelta(days=7)
-    repo.bump_root_by_one()
+    repo.publish([Root.type], bump_version=True)
```
AdamKorcz, line 119: can remove `bump_version=True`.

tuf_conformance/test_expiration.py
```diff
@@ -116,14 +116,14 @@ def test_expired_local_root(client: ClientRunner, server: SimulatorServer) -> No
...
 
     # Refresh
     assert client.refresh(init_data) == 0
 
     # root v3 expires in 21 days
     repo.root.expires = now + datetime.timedelta(days=21)
-    repo.bump_root_by_one()
+    repo.publish([Root.type], bump_version=True)
```
AdamKorcz, line 126: can remove `bump_version=True`.

tuf_conformance/test_file_download.py
```diff
@@ -135,10 +140,9 @@ def test_multiple_changes_to_target(
         new_file_contents = f"new file contents {i}".encode()
         new_target_path = f"new-target-{i}"
         repo.add_artifact(Targets.type, new_file_contents, new_target_path)
-        repo.targets.version += 1
 
         # Bump repo snapshot
```
AdamKorcz, line 144: Could specify that we also bump targets and timestamp.

tuf_conformance/test_file_download.py
```diff
@@ -160,10 +164,9 @@ def test_multiple_changes_to_target(
             # Modify artifact content without updating the hashes/length in metadata.
             malicious_file_contents = f"malicious contents {i}".encode()
             repo.artifacts[target_path].data = malicious_file_contents
-            repo.targets.version += 1
 
             # Bump repo snapshot
```
AdamKorcz, line 168: Could specify that we also bump targets and timestamp.

tuf_conformance/test_keys.py
```diff
@@ -159,17 +156,17 @@ def test_keytype_and_scheme(
...
+    repo.publish([Root.type], bump_version=True)
 
     assert client.refresh(init_data) == 0
     assert client.version(Root.type) == 2
 
     # Create new root version. Replace the correct signature with one that looks
     # reasonable for the keytype but is incorrect. Expect client to refuse the new root.
-    repo.bump_root_by_one()
-    root_md = Metadata.from_bytes(repo.signed_roots.pop())
+    repo.publish([Root.type], bump_version=True)
```
AdamKorcz, line 166: don't need to specify `bump_version=True`

tuf_conformance/test_rollback.py
```diff
@@ -20,12 +20,12 @@ def test_simple_snapshot_rollback(
...
 
     # Repository performs legitimate update to snapshot
-    repo.update_snapshot()  # v2
+    repo.publish([Snapshot.type, Timestamp.type])
     assert client.refresh(init_data) == 0
 
     # Repository attempts rollback attack (note that the snapshot version in
     # timestamp.meta is v2)
-    repo.snapshot.version -= 1  # v1
+    repo.publish([Snapshot.type])
```
AdamKorcz, line 28: Perhaps we can document why this is a rollback.

tuf_conformance/test_rollback.py
```diff
@@ -178,9 +179,12 @@ def test_new_snapshot_version_mismatch(
...
     assert client.init_client(init_data) == 0
 
-    # Increase snapshot version without updating timestamp
-    repo.snapshot.version += 1
+    # Increase snapshot version but do not update version in timestamp.snapshot_meta
+    repo.publish([Snapshot.type])  # v2
+    repo.timestamp.snapshot_meta.version -= 1
+    repo.publish([Timestamp.type])  # v2
 
+    # FIXEM: the expected results are incorrect -- why would the refresh fail?
```
AdamKorcz, line 187: `FIXEM` -> `FIXME`

tuf_conformance/test_rollback.py
```diff
@@ -201,17 +205,20 @@ def test_new_timestamp_fast_forward_recovery(
...
     assert client.refresh(init_data) == 0
-    assert client.version(Timestamp.type) == 99999
+    assert client.version(Timestamp.type) == 100
 
     # repository rotates timestamp keys,
     # rolls back timestamp version
     repo.rotate_keys(Timestamp.type)
-    repo.bump_root_by_one()
+    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
+    repo.publish([Root.type])
```
AdamKorcz, line 219: Combine these two `publish` calls into one.

tuf_conformance/test_basic.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/33284a464e0ae32c93f93da0e8b9c484d552faf7

tuf_conformance/test_file_download.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/abacf0b7b4bd7d2ba7f412802f3d3f10a2718f74

tuf_conformance/test_file_download.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/abacf0b7b4bd7d2ba7f412802f3d3f10a2718f74

tuf_conformance/test_keys.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/487b3882ec4107beece4d9ca2364f3dc8a17a5f4

tuf_conformance/test_rollback.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/a3d010e9c18e6b3891c4d597f638e3245f28cf42

tuf_conformance/test_rollback.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/bde3bcb370feb0fdb431fb10ae3140308e6e60ea

tuf_conformance/test_expiration.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/bf2b7b23b9eaeb0a301361848ef2d9c3df5ff844

tuf_conformance/test_expiration.py
AdamKorcz: https://github.com/theupdateframework/tuf-conformance/pull/188/commits/bf2b7b23b9eaeb0a301361848ef2d9c3df5ff844

tuf_conformance/test_rollback.py
AdamKorcz: @jku PTAL

jku:
>> This does not make a whole lot of sense since if timestamp is not updated ... then how would the client even know to download the new snapshot

> It looks to me like the goal of the test is to ensure that the client respects the snapshot_meta. I have taken a stab at rewriting the test so that it reflects that goal. LMKWYT.

I guess the idea is that a client would be for some reason first download snapshot v1 then v2 then v3 or something? 

But ee don't need to publish a new snapshot to know that the client does not do that: Our first test test_basic_init_and_refresh already checks that there are no unexpected download attempts. : we know for a fact that the client will not download this new snapshot.

I think we can remove this test as well as `test_new_targets_version_mismatch` in test_basic.py

jku (commented):
Cheers, latest changes look good. I have a bunch more (including removal of 3 tests that I cannot make useful. Please push back if there is some client action that these tests could find that existing tests don't: I could not find one)

After my changes we have 0 uses of `bump_version=False`: I suggest we remove it

How about this as a safety check in `publish()`: we verify that the metadata version and "url version" (meaning length of `signed_mds[role]`) match? Then we could add a `verify_version=True` argument to publish so that the tests that do want to have mismatching metadata version and "url version" can still do it -- but it does not happen by accident

**EDIT**: I added the changes I proposed, that even found one inconsistency already

tuf_conformance/test_basic.py
```diff
@@ -160,14 +160,18 @@ def test_timestamp_content_changes(
     assert client.init_client(init_data) == 0
     assert client.refresh(init_data) == 0
 
-    initial_timestamp_meta_ver = repo.timestamp.snapshot_meta.version
-    # Change timestamp without bumping its version in order to test if a new
-    # timestamp with the same version will be persisted.
+    # Change timestamp v1: client should not download a new one if it has v1 already
```
jku, line 163: My comment is incorrect: timestamp is always downloaded, we just expect the client to keep using the version it already has.

jku:
I think we might be ready with this. I will review once more with fresh eyes -- either on friday or monday

jku:
I'm not doing more today but here's two items:
* Example in README needs to be updated
* below is documentation on how this PR changes the test repositories: it would not hurt to have a look through them and verify that we expected these changes



**test repo version numbers, main vs refactor-reposim:**
```
test_basic_metadata_hash_support
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_basic_metadata_hash_support
< {"targets": 2, "snapshot": 3, "timestamp": 3, "root": 1}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 1}

test_client_downloads_expected_file
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_client_downloads_expected_file
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_client_downloads_expected_file_in_sub_dir
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_download_with_hash_algorithms%5Bhashes0%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_download_with_hash_algorithms%5Bhashes1%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_download_with_hash_algorithms%5Bhashes2%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_download_with_unknown_hash_algorithm
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_duplicate_keys_root
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 2}

test_expired_local_timestamp
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 1, "snapshot": 1, "timestamp": 2, "root": 1}

test_expired_local_timestamp
< {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 3, "root": 1}

test_graph_traversal%5Bbasic-delegation%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1}

test_graph_traversal%5Bcyclic-graph%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_graph_traversal%5Bsingle-level-delegations%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "B": 1, "A": 1}

test_graph_traversal%5Bthree-level-delegation-test-DFS-order-of-traversal%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_graph_traversal%5Bthree-level-ignores-all-branches-not-matching-paths%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_graph_traversal%5Bthree-level-terminating-ignores-all-but-roles-descendants%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_graph_traversal%5Btwo-level-delegations%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "B": 1, "A": 1, "C": 1}

test_graph_traversal%5Btwo-level-ignores-all-branches-not-matching-paths%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "B": 1, "A": 1, "C": 1}

test_graph_traversal%5Btwo-level-terminating-ignores-all-but-roles-descendants%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_graph_traversal%5Btwo-level-test-DFS-order-of-traversal%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_graph_traversal%5Btwo-roles-delegating-to-a-third%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "B": 1, "A": 1, "C": 1}

test_graph_traversal%5Btwo-roles-delegating-to-a-third-different-paths%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "B": 1, "A": 1, "C": 1}

test_multiple_changes_to_target
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_multiple_changes_to_target/refresh-10/repo.json /tmp/refactor-reposim-dump/
test_multiple_changes_to_target/refresh-10/repo.json
< {"targets": 15, "snapshot": 15, "timestamp": 15, "root": 1}
> {"targets": 16, "snapshot": 16, "timestamp": 16, "root": 1}

test_multiple_changes_to_target
< {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 1}

test_multiple_changes_to_target
< {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 1}

test_multiple_changes_to_target
< {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 1}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 1}

test_multiple_changes_to_target
< {"targets": 8, "snapshot": 8, "timestamp": 8, "root": 1}
> {"targets": 9, "snapshot": 9, "timestamp": 9, "root": 1}

test_multiple_changes_to_target
< {"targets": 8, "snapshot": 8, "timestamp": 8, "root": 1}
> {"targets": 9, "snapshot": 9, "timestamp": 9, "root": 1}

test_multiple_changes_to_target
< {"targets": 9, "snapshot": 9, "timestamp": 9, "root": 1}
> {"targets": 10, "snapshot": 10, "timestamp": 10, "root": 1}

test_multiple_changes_to_target
< {"targets": 14, "snapshot": 14, "timestamp": 14, "root": 1}
> {"targets": 15, "snapshot": 15, "timestamp": 15, "root": 1}

test_multiple_changes_to_target
< {"targets": 14, "snapshot": 14, "timestamp": 14, "root": 1}
> {"targets": 15, "snapshot": 15, "timestamp": 15, "root": 1}

test_new_snapshot_fast_forward_recovery
< {"targets": 1, "snapshot": 99999, "timestamp": 2, "root": 1}
> {"targets": 1, "snapshot": 100, "timestamp": 2, "root": 1}

test_new_snapshot_fast_forward_recovery
< {"targets": 1, "snapshot": 1, "timestamp": 3, "root": 2}
> {"targets": 2, "timestamp": 3, "root": 2, "snapshot": 1}

test_new_targets_fast_forward_recovery
< {"targets": 99999, "snapshot": 2, "timestamp": 2, "root": 1}
> {"targets": 100, "snapshot": 2, "timestamp": 2, "root": 1}

test_new_targets_fast_forward_recovery
< {"targets": 1, "snapshot": 3, "timestamp": 3, "root": 2}
> {"snapshot": 3, "timestamp": 3, "root": 2, "targets": 1}

test_new_timestamp_fast_forward_recovery
< {"targets": 1, "snapshot": 1, "timestamp": 99999, "root": 1}
> {"targets": 1, "snapshot": 1, "timestamp": 100, "root": 1}

test_new_timestamp_fast_forward_recovery
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 1, "snapshot": 1, "root": 2, "timestamp": 1}

test_new_timestamp_version_rollback
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 1, "snapshot": 1, "root": 1, "timestamp": 1}

test_non_root_rotations%5B1-of-1-key-rotation%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B1-of-1-key-rotation%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B1-of-1-key-rotation%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B1-of-1-key-rotation-fail-not-signed-with-new-key%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B1-of-1-key-rotation-fail-not-signed-with-new-key%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B1-of-1-key-rotation-fail-not-signed-with-new-key%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B1-of-1-key-rotation-unused-signatures%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B1-of-1-key-rotation-unused-signatures%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B1-of-1-key-rotation-unused-signatures%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B2-of-5-one-key-signature-mising-threshold-not-reached%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B2-of-5-one-key-signature-mising-threshold-not-reached%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B2-of-5-one-key-signature-mising-threshold-not-reached%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B3-of-5-one-key-signature-wrong-not-signed-with-3-expected-keys%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B3-of-5-one-key-signature-wrong-not-signed-with-3-expected-keys%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B3-of-5-one-key-signature-wrong-not-signed-with-3-expected-keys%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B3-of-5-sign-fifth-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B3-of-5-sign-fifth-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B3-of-5-sign-fifth-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B3-of-5-sign-first-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B3-of-5-sign-first-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B3-of-5-sign-first-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B3-of-5-sign-fourth-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B3-of-5-sign-fourth-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B3-of-5-sign-fourth-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B3-of-5-sign-second-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B3-of-5-sign-second-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B3-of-5-sign-second-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_non_root_rotations%5B3-of-5-sign-third-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_non_root_rotations%5B3-of-5-sign-third-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 3}

test_non_root_rotations%5B3-of-5-sign-third-combo%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 4, "snapshot": 4, "timestamp": 4, "root": 4}

test_repository_substitutes_target_file
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_repository_substitutes_target_file
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_repository_substitutes_target_file
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_root_expired
< {"targets": 2, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 3}

test_root_rotation%5B1-of-1-key-rotation%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 3}

test_root_rotation%5B1-of-1-key-rotation-fail-not-signed-with-new-key%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B1-of-1-key-rotation-fail-not-signed-with-old-key%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B1-of-1-key-rotation-unused-signatures%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 3}

test_root_rotation%5B1-of-2-threshold-increase-to-2-of-2%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B1-of-3-threshold-bump-to-2-of-3-fails-new-threshold-not-reached%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B1-of-3-threshold-increase-to-2-of-3%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B2-of-3-threshold-decrease-to-1-of-3%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 3}

test_root_rotation%5B2-of-3-threshold-decr-to-1-of-3-fails-old-threshold-not-eached%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B3-of-5-all-keys-rotated-with-intermediate-step%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 3}

test_root_rotation%5B3-of-5-one-key-rotated%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B3-of-5-one-key-rotated-with-intermediate-step%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 3}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 3}

test_root_rotation%5B3-of-5-one-key-rotate-fails-not-signed-with-3-new-keys%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B3-of-5-one-key-rotate-fails-not-signed-with-3-old-keys%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_root_rotation%5B3-of-5-sign-with-different-keycombos%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 4}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 4}
Only in /tmp/main-dump: 
test_simple_snapshot_rollback

test_snapshot_does_not_meet_threshold
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 2}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 2}

test_snapshot_does_not_meet_threshold
< {"targets": 1, "snapshot": 3, "timestamp": 3, "root": 2}
> {"targets": 3, "snapshot": 3, "timestamp": 3, "root": 2}

test_snapshot_rollback%5Bbasic%5D
< {"targets": 1, "snapshot": 1, "timestamp": 3, "root": 1}
> {"targets": 1, "timestamp": 3, "root": 1, "snapshot": 1}

test_snapshot_rollback%5Bwith%20hashes%5D
< {"targets": 1, "snapshot": 1, "timestamp": 3, "root": 1}
> {"targets": 1, "timestamp": 3, "root": 1, "snapshot": 1}

test_targetfile_search%5Bno%20delegations%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_targetfile_search%5Btargetpath%20is%20not%20delegated%20by%20all%20roles%20in%20the%20chain%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_targetfile_search%5Btargetpath%20matches%20wildcard%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_targetfile_search%5Btargetpath%20with%20separators%20x%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_targetfile_search%5Btargetpath%20with%20separators%20y%5D
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1, "A": 1, "B": 1, "C": 1, "D": 1}

test_targets_expired
< {"targets": 1, "snapshot": 2, "timestamp": 2, "root": 1}
> {"targets": 2, "snapshot": 2, "timestamp": 2, "root": 1}

test_targets_rollback%5Bbasic%5D
< {"targets": 1, "snapshot": 3, "timestamp": 3, "root": 1}
> {"snapshot": 3, "timestamp": 3, "root": 1, "targets": 1}

test_targets_rollback%5Bwith%20hashes%5D
< {"targets": 1, "snapshot": 3, "timestamp": 3, "root": 1}
> {"snapshot": 3, "timestamp": 3, "root": 1, "targets": 1}

test_timestamp_content_changes
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 1}
> {"targets": 1, "snapshot": 1, "root": 1, "timestamp": 1}

test_unsigned_metadata%5Broot%5D
< {"targets": 1, "snapshot": 1, "timestamp": 1, "root": 2}
> {"root": 2}

test_unsigned_metadata%5Bsnapshot%5D
< {"targets": 1, "snapshot": 2, "timestamp": 1, "root": 1}
> {"root": 1, "targets": 1, "snapshot": 1, "timestamp": 1}

test_unsigned_metadata%5Btargets%5D
< {"targets": 2, "snapshot": 1, "timestamp": 1, "root": 1}
> {"root": 1, "targets": 1, "snapshot": 1, "timestamp": 1}

test_unsigned_metadata%5Btimestamp%5D
< {"targets": 1, "snapshot": 1, "timestamp": 2, "root": 1}
> {"root": 1, "targets": 1, "snapshot": 1, "timestamp": 1}

```

AdamKorcz:
> * Example in README needs to be updated
> * below is documentation on how this PR changes the test repositories: it would not hurt to have a look through them and verify that we expected these changes

I am on this.

AdamKorcz:
Some tests I am unsure are correct. Updating as I go.

test_duplicate_keys_root: There are a couple of things: AFAICT, the client should not get snapshot v2, but from @jku's list it looks like it does. Also, the assertion `client.version(Snapshot.type) != 1` seems to not be entirely correct. If the assertion returns true, it means that the client _did_ accept the Snapshot upgrade.

jku:
> test_duplicate_keys_root: There are a couple of things: AFAICT, the client should not get snapshot v2, but from @jku's list it looks like it does. Also, the assertion `client.version(Snapshot.type) != 1` seems to not be entirely correct. If the assertion returns true, it means that the client _did_ accept the Snapshot upgrade.

The list I posted is the _repository state_ (not client state) at time of refresh: so snapshot v2 in repository is correct.

You are correct about the client assert I think: there's a single refresh so we likely want to assert `client.version(Root.type) is None`

AdamKorcz:
I have updated the README and read through the tests. Have two comments on the tests here: https://github.com/theupdateframework/tuf-conformance/pull/188#issuecomment-2367717379

README.md
```diff
@@ -83,14 +83,18 @@ def test_targets_version(
     # Initialize our repository: modify targets version and make sure the version is included in snapshot
     init_data, repo = server.new_test(client.test_name)
     repo.targets.version = 7
-    repo.update_snapshot()  # snapshot v2
+    repo.publish([Targets.type, Snapshot.type, Timestamp.type])
```
jku, line 86: this example will fail in `publish()` as the URL version (2) and metadata version (7) do not match. Same thing happens later when v7 is published.

You could change this example to actually test that (url version and metadata version mismatch) as it's simpler than the current correct example would be. Alternatively you must publish the metadata versions at correct URLs.

AdamKorcz: Updated in https://github.com/theupdateframework/tuf-conformance/pull/188/commits/df9b94aae732c4ebbadb18c42a4a6032cefdf646.

AdamKorcz:
> I'm fine either way (current code or this modification)

Let's keep it as is then.

AdamKorcz:
> You are correct about the client assert I think: there's a single refresh so we likely want to assert `client.version(Root.type)` is 1None1, client should not get any version of snapshot

I believe you are right, only that we should assert that `Snapshot` is `None` instead of `Root`. Updated in https://github.com/theupdateframework/tuf-conformance/pull/188/commits/a061768fe76b4cc675e982291204b0fb1c1fcb9a

jku (approved):
I do not like the example in the README: the amount of example code increased by 50%! It's also highly repetitive: the last two very long example sections are almost complete carbon copies.

That said I really want this PR merged so let's go with this. I will try improving the example later.

jku:
I'll do a squash merge: there are so many broken commits on this tree that this will start making bisects tricky if we just merge the branch as is

jku:
Thanks @AdamKorcz this was a big job

Issue #171, opened 2024-08-22
remove skip when possible


jku:
test_updater_delegation_graphs.py has a hack to skip go-tuf currently: this is a reminder to remove that when go-tuf has a new release that we can test

PR #192, opened 2024-09-20
Enable all delegation tests for go-tuf

rdimitrov:
The following PR enables the delegation tests for the go-tuf client removing the existing skip we had so far

Fixes https://github.com/theupdateframework/tuf-conformance/issues/171


jku:
Scratch that. Dependabot updates were just added so this should pass now...

jku:
Thanks

PR #198, opened 2024-09-26
Improve keyid duplication tests

### LINKED ISSUE #147: test_duplicate_keys_root should add duplicate signatures too
the `test_duplicate_keys_root` test is pretty good:
* it adds duplicate keyids (and one matching key) to snapshot role and signs the new root
* test expects client refresh to fail This can happen for two reasons
   * either the client does not accept the new root because of duplicate keyids (#108)
   * or client does  accept new root but snapshot does not meet threshold
 
 
The one thing I'd like to add is that _snapshot metadata should contain duplicate signatures_: the same keyid/signature pair multiple times.

Implementing this requires the RepositorySimulator refactor "explicit signing #155 since we need to manually tweak the signatures .

jku:
* Make test_duplicate_keys actually have a (fake) threshold of signatures: the same keyid is now repeated in both roles keyids and the delegated metadata signatures
* Add a separate test for just duplicated keyids in signatures: this is forbidden by spec but was not always so explicitly forbidden -- this tests a specific spec statement but the older test_duplicate_keys is still useful as it tests the real "attack" of trying to fake the threshold

Fixes #147

### DIFF
--- a/tuf_conformance/test_keys.py
+++ b/tuf_conformance/test_keys.py
@@ -1,3 +1,5 @@
+import json
+
 import pytest
 from tuf.api.metadata import Metadata, Root, Snapshot, Targets, Timestamp
 
@@ -86,7 +88,7 @@ def test_snapshot_has_too_few_keys(
     assert client.version(Snapshot.type) != 2
 
 
-def test_duplicate_keys_root(client: ClientRunner, server: SimulatorServer) -> None:
+def test_duplicate_keys(client: ClientRunner, server: SimulatorServer) -> None:
     """Test multiple identical keyids, try to fake threshold
 
     Client should either not accept metadata with duplicate keyids in a role,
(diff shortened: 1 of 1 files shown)


AdamKorcz (approved):

tuf_conformance/test_keys.py
```diff
@@ -109,9 +111,51 @@ def test_duplicate_keys_root(client: ClientRunner, server: SimulatorServer) -> N
...
+    assert sig is not None
+    for n in range(0, 8):
+        md["signatures"].append(sig)
+    repo.signed_mds[Snapshot.type].append(json.dumps(md).encode())
+
     # This should fail for one of two reasons:
     # 1. client does not accept root v2 metadata that contains duplicate keyids or
-    # 2. client did accept root v2 but then snapshot threshold is not reached
+    # 2. client did accept root v2 but does not accept because it contains duplicate
+    #    keyids (and does not actually meet threshold)
```
jku, line 129: missing the word "snapshot" here, will add

PR #200, opened 2024-09-27
Add test with invalid signatures in valid metadata

### LINKED ISSUE #37: tests: test valid metadata with invalid signature
This is a specific case where folks have interpreted spec in two ways before:

When metadata signatures contains an invalid signature the metadata can still be valid (as long as threshold of valid signatures is reached)

jku:
The specification is not actually 100% clear but this seems like the right choice for maximum compatibility.

Fixes #37

### DIFF
--- a/tuf_conformance/test_keys.py
+++ b/tuf_conformance/test_keys.py
@@ -214,3 +214,48 @@ def test_keytype_and_scheme(
 
     assert client.refresh(init_data) == 1
     assert client.version(Root.type) == 2
+
+
+def test_invalid_sig_in_valid_metadata(
+    client: ClientRunner,
+    server: SimulatorServer,
+) -> None:
+    """Test that client accepts valid metadata if it contains an invalid signature
+
+    Throughout the test timestamp is signed by two keys: one signature is always correct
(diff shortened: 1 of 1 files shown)


tuf_conformance/test_keys.py
```diff
@@ -214,3 +214,48 @@ def test_keytype_and_scheme(
...
+
+
+def test_invalid_sig_in_valid_metadata(
+    client: ClientRunner,
+    server: SimulatorServer,
+) -> None:
+    """Test that client accepts valid metadata if it contains an invalid signature
+
+    Throughout the test timestamp is signed by two keys: one signature is always correct
+    and that is enough to reach threshold. First part uses a plausible looking signature
```
AdamKorcz, line 226: Not clear what "First part" and "second part" refer to.

AdamKorcz (commented):
LGTM. It might just be me who isn't fully clear on "first" and "second part"

tuf_conformance/test_keys.py
jku: cheers, Tweaked docstring to refer to first and second refresh.

AdamKorcz (approved):

PR #208, opened 2024-10-01
Add test for incorrect metadata type

### LINKED ISSUE #180: test wrong metadata type
from @segiddins :

> what if the fetched metadata has the wrong type for the endpoint?

I thought originally this is just another for form of "malformed metadata" but maybe an implementation might deserialize as generic metadata and then forgets to verify type in which case this could be a different issue. A little far fetched but testing is cheap: let's test this, it should be easy after #155

jku:
Try hard to make the metadata otherwise valid, just the type field is incorrect here.

Fixes #180

### DIFF
--- a/tuf_conformance/test_basic.py
+++ b/tuf_conformance/test_basic.py
@@ -1,11 +1,10 @@
-# Test runner
 import json
 import os
 
 import pytest
 from securesystemslib.formats import encode_canonical
 from securesystemslib.hash import digest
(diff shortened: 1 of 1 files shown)


AdamKorcz:
This one looks ready. Do you want to fix the conflict so we can merge, @jku ?

jku:
rebased on main (and removed one useless comment line)

AdamKorcz (approved):

PR #215, opened 2024-10-02
Byte comparison tweaks

### LINKED ISSUE #214: tweak metadata byte-by-byte comparisons
Currently we have tests that literally compare the clients trusted metadata bytes to the bytes repository serves (see e.g. test_non_root_rotations):
* this is useful to test once (clients really should store the exact bytes so that hash verification of local metadata is possible)
* but most of the tests really are trying to verify that the metadata _content_ is correct: they should compare deserialized metadata to deserialized metadata

This way if a client messes up and stores a different serialization, it does not hide the results of the second type of test

jku:
Fixes #214: This should allow testing clients that use their own serialization format when storing trusted metadata

### DIFF
--- a/tuf_conformance/client_runner.py
+++ b/tuf_conformance/client_runner.py
@@ -120,3 +120,23 @@ class ClientRunner:
             roles.append((rolename, md.signed.version))
 
         return roles
+
+    def assert_metadata(self, role: str, expected_bytes: bytes | None) -> None:
+        """Assert that trusted roles metadata matches the expected bytes
+
+        This assert uses deserialized comparison: See test_metadata_bytes_match
+        for the test that requires byte-for-byte equality.
+        """
+        try:
+            trusted = MetadataTest.from_file(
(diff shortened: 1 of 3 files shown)


tuf_conformance/client_runner.py
```diff
@@ -120,3 +120,23 @@ def trusted_roles(self) -> list[tuple[str, int]]:
             roles.append((rolename, md.signed.version))
 
         return roles
+
+    def assert_metadata(self, role: str, expected_bytes: bytes | None) -> None:
```
jku, line 124: this is a little different from trusted_roles() and version() -- as those do not assert, they only return the current value. This felt easier to use correctly in tests but I could change it.

AdamKorcz (approved):

tuf_conformance/client_runner.py
AdamKorcz: I think it is okay for now. I am fine with changing later if we see an issue here.

PR #239, opened 2024-11-22
Improve assert output on failure

### LINKED ISSUE #230: Output diff or expected/actual?
https://github.com/theupdateframework/tuf-conformance/blob/main/tuf_conformance/client_runner.py#L142 is a somewhat difficult error message to act on.

A more helpful failure might be output expected vs actual. We know metadata is serialized as a string in these tests. If it's too much to output to the screen then write it to a results file and upload it as an asset on the job? or like a deepdiff or something?

jku:
* Failures where the assert happened in Clientrunner.assert_metadata where not handled as pytest verbose asserts: Fix this by registering the module for assert rewriting
* assert on serialized metadata so that pytest assert gives us nice output
* Increase verbosity in Makefile so it's at same level as the action

Fixes #230, mostly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ fix: env/pyvenv.cfg
 
 PHONY: test-python-tuf
 test-python-tuf: dev
-	./env/bin/pytest tuf_conformance \
+	./env/bin/pytest -v tuf_conformance \
 		--entrypoint "./env/bin/python ./clients/python-tuf/python_tuf.py" \
 		--repository-dump-dir $(DUMP_DIR)
 	@echo Repository dump in $(DUMP_DIR)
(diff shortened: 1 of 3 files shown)


jku:
@loosebazooka this should help a little.

```
>       assert trusted == b"", f"Unexpected trusted role {role} content"
E       AssertionError: Unexpected trusted role timestamp content
E       assert b'{\n "signat...on": 2\n }\n}' == b''
E         
E         Full diff:
E         - b''
E         + (b'{\n "signatures": [\n  {\n   "keyid": "e04d636e13f11a220ea4142d3950ac50b8ee'
E         +  b'6af91e95effbb3bee4df114d6b4e",\n   "sig": "304402206b30d8e293d3caefb7b81a'
E         +  b'6a5c3ed5cd7ed3a1e15d8df5e0da3cef56fdd6f38f02205c5b130c1403e3fba3babdba373711'
E         +  b'648dab5154c57e8e2fe634cb04698c126e"\n  },\n  {\n   "keyid": "ec9a5050bb6c78'...
E         
E         ...Full output truncated (9 lines hidden), use '-vv' to show

tuf_conformance/client_runner.py:138: AssertionError
```


Note that the pytest verbosity level affects how much diff pytest spits out: 
* at current  verbosity level used by both Makefile and the action we get a handful of lines
* adding "-vv" gets you full diff I think -- but we don't currently expose verbosity level in either makefile or the action, you can only do that by manually invoking pytest

PR #260, opened 2025-01-30
test artifact caching

jku:
Artifact caching is not part of the spec but it is a good idea
* Client CLI document did not mention caching before: Add it in the doc
* Add test that verifies that two download commands do not result in two downloads
* python test client did not support this so support was added

we already have tests for making sure that if artifact content changes the new content is downloaded, so I did not test that here

### DIFF
--- a/tuf_conformance/test_file_download.py
+++ b/tuf_conformance/test_file_download.py
@@ -238,3 +238,28 @@ def test_download_with_unknown_hash_algorithm(
     # and only then realize hash cannot be verified
     assert client.download_target(init_data, target_path) == 1
     assert client.get_downloaded_target_bytes() == []
+
+
+def test_artifact_cache(client: ClientRunner, server: SimulatorServer) -> None:
+    """In this test client is asked to download the same artifact twice.
+    The client is expected return the cached content on the second time.
+
+    Artifact caching is not required in the specification: clients
+    that do not support it can mark this test as expected to fail
+    """
+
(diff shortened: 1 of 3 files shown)


jku:
The potential argument against adding this test is that caching might not be a feature of the tuf library itself (since it's somewhat trivial for most applications to do) but
* I think xfailing the test is easy enough for users that don't see the point here
* If it's easy for apps,it's also easy for the conformance client

tuf_conformance/test_file_download.py
```diff
@@ -238,3 +238,28 @@ def test_download_with_unknown_hash_algorithm(
     # and only then realize hash cannot be verified
     assert client.download_target(init_data, target_path) == 1
     assert client.get_downloaded_target_bytes() == []
+
+
+def test_artifact_cache(client: ClientRunner, server: SimulatorServer) -> None:
```
loosebazooka, line 243: Should we include the test where the artifact should be downloaded again because the contents changed?

jku: `test_multiple_changes_to_target` should handle that along with other cases... 

I can see some advantage from having a simple test that would only test what you suggest (instead of a long test that tests multiple things), but it does not seem strictly necessary 🤷

loosebazooka: Oh no I'm just not aware of what was in there and it was too much work from my phone to go looking. Lgtm.

loosebazooka (approved):

PR #264, opened 2025-02-09
Add sigstore root-signing as static test data

jku:
Add sigstore root-signing repository as static test data

This is now possible since static repos can define a reference time that client-unders-test should use


jku:
> Could this also live in sigstore-conformance? Or do we expect tuf clients to be consuming sigstore metadata outside of the sigstore-clients? Or is the sigstore tuf root a good example of real world tuf use?

I don't think sigstore-conformance needs something like this -- passing that test suite already implicitly requires being able to handle the live root-signing repo at that moment.

I added this here for two reasons:
* I think the repository is a decent compliance test: if your client is spec-compliant it should be able to operate with root-signing
* Even if it turns out there's something a bit wrong with the metadata, it's still a sample of a real world repository

> Can we include a readme on how to keep this up to date? (should it be kept regularly up to date? Does it have an effect on breaking clients?)

I didn't include a "version number" in the directory name like I did for the tuf-on-ci demo content but the way I was thinking about it was that updating is likely not ideal as this is the state of the repository at a specific time: we don't want to stop testing that state even if we did add another one. We can add complete new repositories later (like `sigstore-root-signing-2026`) if there are changes that seem meaningful to test. I don't know if a regularly updating is that useful...

PR #270, opened 2025-02-17
ClientRunner: accept fake time in download_target()

### LINKED ISSUE #269: faketime in static repositories is not working
* I added faketime support in the tests that use static repositories (#264)
* one of those static repositories is now expired
* the faketime is not working somehow, so `test_static_repository[sigstore-root-signing]` fails

jku:
refresh() already supported setting a fake time, add it to download_target as well.

Use this feature in test_static_repository().

Fixes #269


jku:
I'm merging this as main is currently broken without it

PR #306, opened 2025-09-03
Add missing test for snapshot hashes

jku:
There was a test for targets hashes already, this just extends that:
* Have a separate "happy path" test that has correct hashes
* Add a targets snapshot hashes test that is similar to the targets hash test

This was generated with gemini using prompt:
>  this code base is a tuf (the update framework) client conformance test suite, see https://theupdateframework.github.io/specification/latest/ for the specification, especially "detailed client workflow"
> 
> Are there any missing tests?

Not the most earth shattering find and I had to clean the actual implementation afterwards but looks valid.

### DIFF
--- a/tuf_conformance/test_basic.py
+++ b/tuf_conformance/test_basic.py
@@ -239,18 +239,54 @@ def test_basic_metadata_hash_support(
     # Verify client accepts correct hashes
     assert client.refresh(init_data) == 0
 
+
+def test_targets_hash_mismatch(client: ClientRunner, server: SimulatorServer) -> None:
+    """Verify that clients supports hashes for metadata"""
+    init_data, repo = server.new_test(client.test_name)
+
+    # Construct repository with hashes in timestamp/snapshot
+    repo.compute_metafile_hashes_length = True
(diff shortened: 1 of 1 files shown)


tuf_conformance/test_basic.py
```diff
@@ -239,18 +239,54 @@ def test_basic_metadata_hash_support(
...
-        "46419349341cfb2d95f6ae3d4cd5c3d3dd7f4673985dad42a45130be5e0531a0"
-    )
+    repo.snapshot.meta["targets.json"].hashes["sha256"] = "0" * 64
     repo.publish([Snapshot.type, Timestamp.type])  # v3
 
     # Verify client refuses targets v3 that does not match hashes
     assert client.refresh(init_data) == 1
     assert client.version(Snapshot.type) == 3
-    assert client.version(Targets.type) == 2
+    assert client.version(Targets.type) != 3
```
jku, line 263: using negative test here to allow clients to delete the (no longer valid) metadata if they want to: we're just confirming client does not update to targets v3

PR #389, opened 2026-06-03
Use uv

jku:
Use uv:
* makes installs a bit faster
* Everything is pinned with with uv.lock: the action installs with "--frozen" so only uses the documented uv.lock contents
* pyproject.toml now does not pin any version numbers: dependabot only updates uv.lock
* Makefile now uses "uv run"
* the python selftest client has its own virtualenv (but uses same lock file to make updates management less work)

This does mean uv is a "system dependency" like python -- we expect developers to have it. It might be annoying to some people


.github/workflows/ci.yml
```diff
@@ -20,11 +20,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
-      - name: Set up Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+      - name: Set up uv
+        uses: astral-sh/setup-uv@1edb52594c857e2b5b13128931090f0640537287 # v5.3.0
         with:
           python-version: "3.14"
```
jku: I think I'd rather have a .python-version in repo

.github/workflows/ci.yml
```diff
@@ -20,14 +20,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
 
-      - name: Set up Python
-        uses: actions/setup-python@ece7cb06caefa5fff74198d8649806c4678c61a1 # v6.3.0
+      - name: Set up uv
+        uses: astral-sh/setup-uv@1edb52594c857e2b5b13128931090f0640537287 # v5.3.0
```
jku: this version is really old, should maybe bump that before merge

clients/python-tuf/python_tuf.py
```diff
@@ -1,4 +1,13 @@
-#!/usr/bin/env python
+#!/usr/bin/env -S uv run --script
+#
+# /// script
+# dependencies = [
+#     "tuf>=6.0",
+#     "securesystemslib[crypto]>=1.4.0",
+# ]
+# ///
```
jku: I like this as a really simple way to separate the selftest env from the actual test suite env... It does mean that selftest versions are not strictly pinned.

I have another approach where we could still use the pinned versions from uv.lock

jku: I decided to pin the versions -- this means bringing back a it of manual venv management but we can still use the same uv.lock (as long as our tuf/securesystemslib versions are compatible)

PR #398, opened 2026-07-01
Test ML-DSA keytype

jku:
Added ML-DSA keys as defined in https://github.com/theupdateframework/taps/pull/195 and some additional reasonable ECDSA and RSA schemes: This currently uses securesystemslib main branch but is ready for review.

* Current `test_keytype_and_scheme` is preserved for the "standard" ketytypes (the ones referenced in the spec), `test_mldsa_keytype_and_scheme` is added for ML-DSA. This way other more optional keytypes will be ergonomic for users to xfail when needed.
* Add tests for additional RSA and ECDSA keys: 
  * Test rsassa-pss-sha384 & rsassa-pss-sha512  (don't test for pkcs1v15 as a legacy padding scheme
  * Test ecdsa-sha2-nistp384 (don't test for ecdsa-sha2-nistp521 as it is not commonly used) 
* Refactor the keytype test so we don't have huge keys in source code
* Use securesystemslib main branch (ML-DSA support is merged but not released)
* Enable ML-DSA verification in test suite and python selftest client  (it's not yet enabled by default)
* Fix an unrelated issue where python selftest client fails to find the xfails file when run from Makefile

Partially AI generated.


tuf_conformance/test_keys.py
```diff
@@ -188,13 +166,69 @@ def test_keytype_and_scheme(
...
+
+@pytest.mark.parametrize("keytype, scheme", mldsa_params)
+def test_mldsa_keytype_and_scheme(
+    client: ClientRunner,
+    server: SimulatorServer,
+    keytype: str,
+    scheme: str,
+    enable_mldsa: None,
+) -> None:
+    """Test that client supports ML-DSA keytypes"""
```
jku: maybe I should expand this a bit with details

clients/python-tuf/python_tuf.py
```diff
@@ -10,8 +10,18 @@
 import shutil
 import sys
 
+from securesystemslib.signer import KEY_FOR_TYPE_AND_SCHEME, SSlibKey
 from tuf.ngclient import Updater
 
+# ML-DSA is not enable yet: enable to test the test suite
```
loosebazooka: this comment is a bit confusing as I can't tell how it's currently disabled.

tuf_conformance/conftest.py
```diff
@@ -98,9 +98,7 @@ def read_xfails(pytestconfig: pytest.Config) -> list[str]:
 
 
 @pytest.fixture(autouse=True)
-def conformance_xfail(
-    pytestconfig: pytest.Config, request: pytest.FixtureRequest
-) -> None:
+def conformance_(pytestconfig: pytest.Config, request: pytest.FixtureRequest) -> None:
```
loosebazooka: this seems like an accidental rename? Or I'm missing something?

loosebazooka (commented):
I was going through this and was wordering if it made sense to split the refactoring and classical key additions into their own PR then add mldsa stuff?

If not I'll continue with a review here.

clients/python-tuf/python_tuf.py
jku: yeah, it's not enabled in _securesystemslib_ by default yet, I'll add that in the comment

tuf_conformance/conftest.py
jku: oh you're right. It kept working because the it happens to be an autouse fixture that gets enabled for every test but the rename is an accident

